PR #130, opened 2022-07-26
Fix typo in ownership example

toriningen:
The existing example is incorrect, since it just shadows the `x` with undefined `y`:

```rust
use core::sync::atomic::AtomicI32;

let x: AtomicI32 = AtomicI32::new(42);
let x: AtomicI32 = y; // d-oh!
```

This tiny PR changes the last line to be correct and match textual description:

```rust
let y: AtomicI32 = x;
```


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #131, opened 2022-07-26
Fix core::ops::Drop missing colon typo

toriningen:
In few places `core::ops::Drop` trait is incorrectly referenced to as `core::ops:Drop`, with single colon. This PR fixes that typo.


pietroalbini (approved):
Thanks for spotting this!

pietroalbini:
bors merge

PR #165, opened 2022-07-26
Introduce and reference "C" term

### LINKED ISSUE #159: "C" is never referenced
There should probably be a reference to the ISO C standard, similar to the Unicode one.

ojeda:
I used an explicit edition (current latest), considering the Unicode glossary term also does that.

Closes #159.


kirtchev-adacore:
Thank you for your PR, much appreciated!

kirtchev-adacore:
bors merge

ojeda:
My pleasure!

PR #169, opened 2022-07-27
Fix the patterns chapter not referencing unit enum variants and unit structs

Veykril:
These are separately listed from constants in the specification, but are missing in this chapter.
Fixes https://github.com/ferrocene/specification/issues/132


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #174, opened 2022-07-27
Add internationalization support

ZhangHanDong:
Added internationalization support for language specifications to facilitate translation into other languages such as Chinese, Japanese, German, etc.

## Environmental Installation

```shell
>  pip install sphinx-intl
```

## Add the language you want to internationalize

1. Generate pot in the build/locale directory:

```shell
> sphinx-build -b gettext src build/locale
```

The `.pot` files will be generated under `build/locale`.

2. Generate internationalized country code:

```shell
> sphinx-intl update -p build/locale -l zh_CN
```

The `.po` files will be generated under `locales/zh_CN/LC_MESSAGES/` (Configure in the `src/conf.py` file).

3. Find the corresponding `.po' file and start translating：

```shell
msgid "Ferrocene Language Specification"
msgstr "Ferrocene 语言规范"
```

Modify only the `msgstr` part.

4. Open `src/conf.py` and under the `locale_dirs` configuration, add the following:

```shell
language="zh_CN"
```

5. Recompile.

```shell
> ./make.py
```

The document generated in `build/html` then corresponds to the language you specify, such as `zh_CN`.


pietroalbini:
Thanks for your interest in the Ferrocene Language Specification!

Right now we just published the first draft, and we're planning major revisions over the coming months as we approach our goal of finalizing the specification by the end of the year. I don't think it's worth to start translating right now, with some chapters planned to have full rewrites.

Once the specification is finished we welcome you to open an issue to discuss possible translations: we'll need to choose the right approach, but also ensure our team has the capacity to review translation PRs (unfortunately right now we don't have that, see [the contributions guidelines](https://github.com/ferrocene/specification/blob/main/CONTRIBUTING.rst)).

ZhangHanDong:
@pietroalbini Thanks for your reply.
Now it's not about translating content, it's about adding internationalization skills so that the content is ready for translation

pietroalbini:
To be clear, once the specification is more stable it'd be nice to support translations! I think it'd be better to add support internationalization once we decide to do that though. Again, thanks for your interest in the FLS!

ZhangHanDong:
@pietroalbini Good

skade:
As the person who maintained the translation of the rust-lang.org website, it’s definitely by our values to provide translation. It’s my experience though that I prefer to have that planned (including support) and rigorously executed with people on both sides available for at least a bit of time. A feature not used may bitrot.

Thanks for showing how enabling could work!

ZhangHanDong:
@skade 

I absolutely agree with you.I also organized, translated and reviewed the Chinese page of rust-lang.org.

Looking forward to the stability of the specification.

Issue #58, opened 2022-07-25
Fix use of "binding", "local variable", "object", "temporary", "variable"


pietroalbini:
*Issue exported from our legacy tracking system.*

crlf0710:
Specially, the concept `object` is especially concerning, for example this code compiles today:
```rust
fn main() {
    let x: ();
}
```
There's a local variable that doesn't corresponds to a value at all. `Place`-related terms might explain the thing in a better way.

kirtchev-adacore:
Thank you for the input!

Indeed, `object` is not a suitable term for Rust, and we will drop it.
Constants, statics, and variables (including bindings, local, local variables, and other similar aliases) will be defined in terms of values (or the lack there of).

crlf0710:
Constants are defined in terms of values, statics are places always containing values. \<insert something about thread locals here.\> Variables are named places that sometimes has values if a value has been put into it, and the value hasn't been moved out of. 

Also note that in restricted cases it is also possible to move value out of the inner place provided by an outer value, e.g. moving a single field value out of a struct value, then the outer value becomes "incomplete" or "destructured". This is only possible when it doesn't have `Drop` implemented.

Something like this.

PR #176, opened 2022-07-27
Fix use of "binding", "local variable", "object", "temporary", "variable"

kirtchev-adacore:
This PR updates the glossary in the following manner:

- Removes the term "object".
- Updates the definition of terms "binding", "constant", "local", "local variable", "static", "temporary", and "variable".

As a consequence, the PR also updates the uses of the aforementioned terms in relevant chapters.

Fixes [#58](https://github.com/ferrocene/specification/issues/58)

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -3436,7 +3436,7 @@ Method Call Expressions
 
 :dp:`fls_b7i26954j1hc`
 A :t:`method call expression` is an :t:`expression` that invokes a :t:`method`
-of an :t:`object`.
+of a :t:`variable`.
 
 :dp:`fls_jx3ryre0xs88`
 A :t:`receiver operand` is an :t:`operand` that denotes the :t:`value` whose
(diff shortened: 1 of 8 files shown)


Veykril (dismissed):

src/ownership-and-deconstruction.rst
```diff
@@ -20,7 +20,7 @@ Ownership
 management model of Rust.
 
 :dp:`fls_ckcnkbb6y3cq`
-An :t:`owner` is an :t:`object` that holds a :t:`value`.
+An :t:`owner` is a :t:`variable` that holds a :t:`value`.
```
Veykril, line 23: This should state `value holder` instead of variable actually I think. A `static` can very much own a value (it's just never dropped), a const technically as well (though there given its usages is basically placed at the use site it doesn't really use ownership actively).

We might wanna split that off into an issue though, not sure how the chapter talks about borrowing in regards to the term `owner` (since statics are a bit different in regards to borrowing, mutable borrows needing unsafe etc)

Veykril (approved):

src/ownership-and-deconstruction.rst
kirtchev-adacore: Will use `variable holder` and spawn off a separate issue to revisit `variable holder` and `owner`.

Veykril (approved):

Veykril:
bors merge

src/expressions.rst
```diff
@@ -375,7 +375,7 @@ The following :t:`[expression]s` are :t:`[place expression]s`:
   expression`,
 
 * :dp:`fls_ya05djl1d154`
-  :t:`[Path expression]s` that resolve to a :t:`binding` or a :t:`static`.
+  :t:`[Path expression]s` that resolve to a :t:`static` or a :t:`variable`.
```
steveklabnik, line 378: Glad to see this change; long ago we discussed this endlessly for the tutorial and then book, and decided on 'variable' there too 👍

PR #182, opened 2022-07-29
Remove static errors from dynamic semantics in pattern matching

Veykril:
These belong into the legality rules of the corresponding entities.
They will be reintroduced by the corresponding subchapter rewrites.
Fixes https://github.com/ferrocene/specification/issues/137


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #184, opened 2022-07-29
Specify that arithmetic overflow is not undefined behavior

Veykril:
Fixes https://github.com/ferrocene/specification/issues/143

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -337,8 +337,7 @@ A :t:`constant expression` is evaluated statically whenever its :t:`value` is
 needed.
 
 :dp:`fls_b46nyamfqxdu`
-The :t:`evaluation` of a :t:`constant expression` that results in arithmetic
-overflow :t:`[panic]s`.
+The :t:`evaluation` of a :t:`constant expression` that results in :t:`arithmetic overflow` :t:`[panic]s`.
 
 :dp:`fls_ms9vey2wymqp`
 It is a static error if a :t:`constant expression` either :t:`[panic]s` or
@@ -1309,11 +1308,9 @@ The :t:`evaluation` of a :t:`subtraction expression` proceeds as follows:
 #. :dp:`fls_b9g0r9vc4rou`
    ``core::ops::Rem::rem(left_operand, right_operand)`` is invoked.
 
-.. rubric:: Undefined Behavior
-
-:dp:`fls_8dkygceg0oo`
-It is undefined behavior for an :t:`arithmetic operation` to cause overflow with
-:t:`[value]s` of :t:`[numeric type]s`.
+:dp:`fls_35oSMqAMFYWl`
+   If :t:`arithmetic overflow` occurs, the computed :t:`value` shall wrap around or the program
+   shall either panic or abort execution.
 
 .. rubric:: Examples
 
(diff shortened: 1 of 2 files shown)


src/expressions.rst
```diff
@@ -1309,11 +1308,9 @@ The :t:`evaluation` of a :t:`subtraction expression` proceeds as follows:
 #. :dp:`fls_b9g0r9vc4rou`
    ``core::ops::Rem::rem(left_operand, right_operand)`` is invoked.
 
-.. rubric:: Undefined Behavior
-
-:dp:`fls_8dkygceg0oo`
-It is undefined behavior for an :t:`arithmetic operation` to cause overflow with
-:t:`[value]s` of :t:`[numeric type]s`.
+:dp:`fls_35oSMqAMFYWl`
+   If :t:`arithmetic overflow` occurs, the computed :t:`value` shall wrap around or the program
```
kirtchev-adacore, line 1312: There are spaces at the start of the sentence.

src/expressions.rst
```diff
@@ -2460,14 +2457,6 @@ assigned_operand, modifying_operand)`` is invoked.
 For a :t:`subtraction assignment`, ``core::ops::SubAssign::sub_assign(&mut
 assigned_operand, modifying_operand)`` is invoked.
 
-.. rubric:: Undefined Behavior
-
-:dp:`fls_uywamh3nzl6p`
-It is undefined behavior for an :t:`addition assignment`, a :t:`division
```
kirtchev-adacore, line 2466: I am not sure that this should be removed, unless these do not cause overflow. Arithmetic expressions use different `core` functions, so we can't argue that `+=` overflows just because it contains `+`.

kirtchev-adacore (changes requested):

src/expressions.rst
```diff
@@ -2027,7 +2024,7 @@ The :t:`evaluation` of a :dt:`numeric cast` proceeds as follows:
 
 * :dp:`fls_2etd73f8jg2n`
   Produces :c:`f32::INFINITY` of the same sign as the :t:`value` of the
-  :t:`operand` when the :t:`value` of the :t:`operand` causes overflow.
+  :t:`operand` when the :t:`value` of the :t:`operand` causes :t:`arithmetic overflow`.
```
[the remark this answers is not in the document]
Veykril: what indent do you mean here? I don't quite follow

src/expressions.rst
Veykril: These are by definition not undefined behaviour, if it is a user implementation, it just does what the user has implemented them to do. If this is a builtin implementation (that is arithmetic on primitives), it is also never undefined behaviour, it will either wrap, panic or abort.

src/expressions.rst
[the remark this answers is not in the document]
kirtchev-adacore: Ignore my claim. When the browser window is resized, the text gets wrapped.

src/expressions.rst
kirtchev-adacore: Understood.

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

steveklabnik:
Not that this is incorrect, but I'm curious as to why this doesn't specify the "debug assertions" requirement with regard to these semantics. Should I open another issue?

Veykril:
Ah, I meant to make a note about that somewhere, but looks like I forgot, thanks for checking!
I haven't addressed it here because iirc we don't have a notion of debug assertions (yet). If you could open an issue regarding the debug assertions requirement that would be appreciated :)

PR #197, opened 2022-08-03
Fix make.py not working properly on windows setups

Veykril:
On windows python installs, venv will contain a `scripts` folder with the executables, instead of a `bin` folder


japaric (approved):

Veykril:
bors merge

PR #198, opened 2022-08-03
Adjust array indexing expression to be more general

### LINKED ISSUE #175: Array index expression is not only for array or a slice.
Paragraph 6.8.2:1

> An array index expression is an expression that indexes into an array or a slice.

Some paragraphs below it is stated that the operand can be of any type implementing the `core::ops::Index` trait.

It should be this (or similar)

> An array index expression is an expression that indexes into an indexed array operand

There are other places where *"array or a slice"* is used, while *"indexed array operand"* would be more correct.

Veykril:
Currently the subchapter is talking very specifically about arrays and slices
but the mechanism works generally via the Index traits

This also removes the `6.8 Array and Array Index Expressions` section, as grouping these two expressions is of no value.

Closes #175

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -30,13 +30,13 @@ Expressions
    ExpressionWithoutBlock ::=
        OuterAttributeOrDoc* (
            ArrayExpression
-         | Array
          | AwaitExpression
          | BreakExpression
          | CallExpression
          | ClosureExpression
          | ContinueExpression
          | FieldAccessExpression
+         | IndexExpression
          | LiteralExpression
          | MethodCallExpression
          | MacroInvocation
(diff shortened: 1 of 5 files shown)


src/expressions.rst
```diff
@@ -2704,78 +2699,77 @@ Array and Slice Indexing Expressions
 .. rubric:: Legality Rules
 
 :dp:`fls_42ijvuqqqlvh`
-An :t:`array index expression` is an :t:`expression` that indexes into an
-:t:`array` or a :t:`slice`.
+An :t:`index expression` is an :t:`expression` that indexes into a :t:`value`
+of a :t:`type`.
```
kirtchev-adacore, line 2703: I wonder whether we should introduce the concept of an "indexable type", something along the lines of "An indexable type is a type that implements the core::ops::Index trait.". I don't think that the phrasing "... indexes into a value of a type" is entirely correct, as theoretically I can index into an integer.

src/expressions.rst
```diff
@@ -2704,78 +2699,77 @@ Array and Slice Indexing Expressions
...
 
 :dp:`fls_pc0c22asgzvw`
-An :t:`indexed array operand` is an :t:`operand` which indicates the :t:`array`
-or :t:`slice` being indexed into by an :t:`array index expression`.
+An :t:`indexed operand` is an :t:`operand` which indicates the :t:`value` of a :t:`type` implementing
+:std:`core::ops::Index` being indexed into by an :t:`index expression`.
```
kirtchev-adacore, line 2707: Assuming that we define what an "indexable type" is, then we can rephrase this as

An indexed operand is an operand which indicates the value being indexed into by an index expression.

The type of an indexed operand shall be an indexable type.

src/expressions.rst
```diff
@@ -2704,78 +2699,77 @@ Array and Slice Indexing Expressions
...
 
 :dp:`fls_ff3sgpldn52o`
-An :t:`indexing operand` is an :t:`operand` which specifies the index of the
-:t:`array` or :t:`slice` being indexed into by an :t:`array index expression`.
+An :t:`indexing operand` is an :t:`operand` which specifies the index for the :t:`indexed operand`
+being indexed into by an :t:`index expression`.
```
kirtchev-adacore, line 2711: Let's relax this to

An indexing operand is an operand which specifies the index of an index expression.

src/expressions.rst
```diff
@@ -4834,7 +4828,7 @@ as follows:
      -
      - left-to-right
    * - :dp:`fls_k3ohh8k888c`
-     - :t:`Array index expression`
+     - :t:`index expression`
```
kirtchev-adacore, line 4831: Alphabetize :D

kirtchev-adacore (changes requested):

src/expressions.rst
Veykril: > I don't think that the phrasing "... indexes into a value of a type" is entirely correct, as theoretically I can index into an integer.

Why is that not correct? A value of a type also applies to an integer
And yes I agree, a new term for indexable type might be good to add

kirtchev-adacore: I was arguing that "type" is not accurate enough, it should be "indexable type" as you can't index into an arbitrary type.

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #224, opened 2022-08-04
Resolve broken links and other minor issues in Chapter 4

### LINKED ISSUE #205: 4. Resolve broken links and other minor issues

kirtchev-adacore:
Fixes #205


jonas-schievink:
bors try

src/glossary.rst
```diff
@@ -443,13 +443,21 @@ An :dt:`attribute macro` is a :t:`procedural macro` that consumes two streams
 of :t:`[token]s` to produce a stream of tokens, and defines a new :t:`outer
 attribute` that can be attached to :t:`[item]s`.
 
+.. _fls_24iVIlHhvnVO:
+
+auto trait
+^^^^^^^^^^
+
+:dp:`fls_d84nTOR4pZq5`
+An :dt:`auto trait` is ??? HELP LUKAS!
```
Veykril, line 452: Auto traits are traits that are automatically implemented for every type, unless the type in question contains at least one type that does not implement the trait or if the type opts-out of the implementation of the trait. (This opt-out is unstable syntax via `impl !Foo for Type {}`, which the std/core lib makes use of.

Basically this part here https://doc.rust-lang.org/beta/unstable-book/language-features/auto-traits.html#automatic-trait-implementations

src/glossary.rst
```diff
@@ -2134,6 +2180,14 @@ generic trait
 :dp:`fls_h515f11akr91`
 A :dt:`generic trait` is a :t:`trait` with :t:`[generic parameter]s`.
 
+.. _fls_3Ss6jDgtF1of:
+
+generic type
+^^^^^^^^^^^^
+
+:dp:`fls_Zn2pIsMZoTry`
+A :dt:`generic type` is ??? HELP LUKAS!
```
Veykril, line 2189: Judging from the sole usage of the term in Variance this seems to be just `a type with a generic parameters`

src/glossary.rst
```diff
@@ -2800,6 +2854,14 @@ defines the compile- and run-time semantics of a program.
 :dp:`fls_xd997kd2i73a`
 See :s:`Item`.
 
+.. _fls_2yPpT2kPT5ys:
+
+item signature
+^^^^^^^^^^^^^^
+
+:dp:`fls_jZ7vlrcEDDLg`
+An :dt:`item signature` is ??? HELP LUKAS!
```
Veykril, line 2863: This is solely used by `Inferred Type`. I think we might want to rewrite the legaility rule there as item signature is not such a good term after all

src/glossary.rst
```diff
@@ -3627,6 +3697,14 @@ An :dt:`operator expression` is an :t:`expression` that involves an operator.
 :dp:`fls_qdszbyeuo7w1`
 See :s:`OperatorExpression`.
 
+.. _fls_C5DiCsvsaBsj:
+
+opt-out trait bound
+^^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_wS4EzN0N1GDP`
+An :dt:`opt-out trait bound` is ??? HELP LUKAS!
```
Veykril, line 3706: An opt-out trait bound is a `?Sized` thing (the question mark bounds).
So something like:
> An opt-out trait bound is a trait bound that allows opting out of implicity introduced trait bounds.

src/glossary.rst
```diff
@@ -5535,6 +5671,14 @@ union value
 :dp:`fls_9BPrxky3a4nE`
 A :dt:`union value` is a :t:`value` of a :t:`union type`.
 
+.. _fls_Is9hWLC6Q0g5:
+
+unique immutable borrow
+^^^^^^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_eXrivAmNxzmv`
+HELP LUKAS!
```
Veykril, line 5680: Though one :) https://doc.rust-lang.org/stable/reference/types/closure.html#unique-immutable-borrows-in-captures

src/glossary.rst
```diff
@@ -5848,6 +6000,26 @@ parameter]s` that describes the circumstances under which a :t:`generic type`
...
+:dp:`fls_z9D86gBFbKB5`
+See :s:`VariadicPart`.
+
+.. _fls_AYmsHgbNcW8r:
+
+variant conformant
+^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_MfiPifTsTzgM`
+HELP LUKAS!
```
Veykril, line 6021: The usage of this term in types and trait is wrong, I think this should be variance conformant.

src/glossary.rst
kirtchev-adacore: Now defined.

src/glossary.rst
kirtchev-adacore: Now defined.

src/glossary.rst
kirtchev-adacore: Now defined.

src/glossary.rst
kirtchev-adacore: Removed.

src/glossary.rst
Veykril: I have removed this definition and expanded the legality rule of `Inferred Type`, please have a look at a856074

src/glossary.rst
Veykril: Added a small definition of this in 0a5bbdd

Veykril (approved):

kirtchev-adacore:
bors merge

PR #228, opened 2022-08-05
Throw an exception if running sphinx-autobuild returns a non-zero exit code

Veykril:


kirtchev-adacore (approved):
I trust your python expertise :D

kirtchev-adacore:
bors merge

make.py
```diff
@@ -34,7 +34,7 @@ def build_docs(root, env, builder, clear, serve):
                 dest / builder,
             ],
             check=False,
-        )
+        ).check_returncode()
```
pietroalbini, line 37: The `check_returncode()` method call can be replaced with `check=True` above.

PR #233, opened 2022-08-09
Fix incorrectly transformed code block in the macro chapter

Veykril:


src/macros.rst
```diff
@@ -980,49 +980,25 @@ When transcribing a metavariable
 A metavariable indication in a macro repetition in transcription shall be
 transcribed to the matched tokens in order,
 
-:dp:`fls_u2lq0lr12kdt`
```
kirtchev-adacore, line 983: There is a bunch of nonsense on lines 976 and 977 which should be removed (our attempts to formulate a rule).

src/macros.rst
```diff
@@ -980,49 +980,25 @@ When transcribing a metavariable
 A metavariable indication in a macro repetition in transcription shall be
 transcribed to the matched tokens in order,
```
kirtchev-adacore, line 981: I feel that the sentence is unfinished. Perhaps this will make it read better:

A metavariable indication in a macro repetition in transcription shall be transcribed to the matched tokens in order, **as follows**:

src/macros.rst
```diff
@@ -980,49 +980,25 @@ When transcribing a metavariable
...
-0;1;2;
+   foo! {
+     0
+     1
+     2
+   }
 
-:dp:`fls_xu5esg3v2u6i`
-Given a repetition in a macro invocation of the form
+   0;1;2;
```
kirtchev-adacore, line 1001: Perhaps this should be preceded with something along the lines of

The macro invocation of foo yields

kirtchev-adacore (changes requested):

kirtchev-adacore (approved):

PR #235, opened 2022-08-09
Describe derive helper attributes

Veykril:
Fixes https://github.com/ferrocene/specification/issues/124

### DIFF
--- a/src/macros.rst
+++ b/src/macros.rst
@@ -408,8 +408,13 @@ The sole parameter of the :t:`macro implementation function` captures
 the :t:`token` stream produced from the related :s:`EnumDeclaration`,
 :s:`StructDeclaration`, or :s:`UnionDeclaration`.
 
+:dp:`fls_H5ipqqlH3pJh`
+A :t:`derive macro` adds all its declared :t:`[derive helper attribute]s` into
+the :t:`abstract data type scope` of the :t:`abstract data type` the
+:t:`attribute` is attached to.
+
 :dp:`fls_mobky5ck1mi`
-A :dt:`helper attribute` is an :t:`inert` :t:`attribute` that acts as a hint to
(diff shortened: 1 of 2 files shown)


src/macros.rst
```diff
@@ -408,8 +408,13 @@ The sole parameter of the :t:`macro implementation function` captures
 the :t:`token` stream produced from the related :s:`EnumDeclaration`,
 :s:`StructDeclaration`, or :s:`UnionDeclaration`.
 
+:dp:`fls_H5ipqqlH3pJh`
+A :t:`derive macro` adds all its declared :t:`[derive helper attribute]s` into
+the :t:`abstract data type scope` of the :t:`abstract data type` the
```
kirtchev-adacore, line 413: Looking at 14.4 Scopes, we do not have the notion of an `abstract data type scope`.

Also, isn't attribute `proc_macro_derive` that adds its derive helper attributes to the abstract data type, and not the derive macro itself?

kirtchev-adacore (changes requested):

src/macros.rst
Veykril: My assumption was that we are planning on redoing the definitions of scope, in which case that would be a thing. I don't think with the current scope definitions this can be properly defined.

`proc_macro_derive` declares the helper attributes of a derive macro. Applying a derive macro then inserts its derive helpers into the scope of the item's definition for attributes. 

```rs
#[derive(MyDerive)]
struct Foo {
    #[MyDeriveHelper]
    field: ()
}
```

Otherwise the use of `MyDeriveHelper` would be an error here, since it can't be resolved as its not in scope.

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #245, opened 2022-08-15
Fixup macro hygiene subchapter

Veykril:


src/macros.rst
```diff
@@ -1044,37 +1044,33 @@ environment.
 Hygiene is categorized as follows:
```
kirtchev-adacore, line 1044: "Hygiene" needs to be defined in the Glossary, and all mentions to the term converted to references.

src/macros.rst
```diff
@@ -1044,37 +1044,33 @@ environment.
...
 
 * :dp:`fls_dz2mvodl818d`
-  *Definition site hygiene*, which resolves to the ``MacroDeclaration``
+  :dt:`Definition site hygiene`, which resolves to the ``MacroDeclaration``
```
kirtchev-adacore, line 1047: `MacroDeclaration` should be syntactic category `MacroRulesDeclaration`.

src/macros.rst
```diff
@@ -1044,37 +1044,33 @@ environment.
...
   site. ``Identifier``\ s with definition site hygiene cannot reference
```
kirtchev-adacore, line 1048: Shouldn't `Identifier` actually be `Path`? Wither way, it is missing the :s: tag.

Also, `definition site hygiene` should reference the term.

src/macros.rst
```diff
@@ -1044,37 +1044,33 @@ environment.
...
   the environment of the ``MacroDeclaration``, cannot be referenced by the
   environment of a ``MacroInvocation``, and are considered *hygienic*.
```
kirtchev-adacore, line 1050: `Hygienic` should be a term.

kirtchev-adacore (changes requested):
Note that there are multiple uses of terms that should be tagged as references of the appropriate kind.

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #249, opened 2022-08-16
Fail CI if glossary.rst does not contain sorted headers

Veykril:
Closes https://github.com/ferrocene/specification/issues/226


pietroalbini (changes requested):
The permissions denied error on CI happened because the file is not executable (`chmod +x`).

Veykril:
Superceded by #258

PR #252, opened 2022-08-16
Specify what a path expression may refer to

Veykril:
Closes https://github.com/ferrocene/specification/issues/196

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -613,6 +613,11 @@ Path Expressions
 :dp:`fls_gvanx4874ycy`
 A :t:`path expression` is an :t:`expression` that denotes a :t:`path`.
 
+:dp:`fls_EOkrcIj9CuhV`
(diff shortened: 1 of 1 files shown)


src/expressions.rst
```diff
@@ -624,11 +629,6 @@ it resolved to.
 :dp:`fls_5ifai8nkp5ek`
 The :t:`value` of a :t:`path expression` is the :t:`entity` that it resolved to.
 
-.. rubric:: Dynamic Semantics
-
-:dp:`fls_ed9w4jwx059`
-The :t:`evaluation` of a :t:`path expression` has no effect.
-
```
Veykril, line 631: This is not true, the evaluation of this may move out of/copy the referenced thing, this is already specified in the place expression rules though.

kirtchev-adacore: Gotcha.

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #254, opened 2022-08-16
Add missing legality rules for closure parameters

Veykril:
Closes https://github.com/ferrocene/specification/issues/106

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -3431,6 +3431,14 @@ A :t:`closure body` denotes a new :t:`control flow boundary`.
 :dp:`fls_fg8lx0yyt6oq`
 A :t:`closure body` is subject to :t:`capturing`.
 
+:dp:`fls_c3rzwUxjmBMY`
+A :t:`closure parameter` is a :t:`construct` that matches an input :t:`value`
+at the site of a :t:`call expression` or a :t:`method call expression` to
+a pattern.
+
+:dp:`fls_r6gWLoNR7JMR`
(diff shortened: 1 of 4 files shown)


src/expressions.rst
```diff
@@ -3431,6 +3431,14 @@ A :t:`closure body` denotes a new :t:`control flow boundary`.
 :dp:`fls_fg8lx0yyt6oq`
 A :t:`closure body` is subject to :t:`capturing`.
 
+:dp:`fls_c3rzwUxjmBMY`
+A :t:`closure parameter` is a :t:`construct` that matches an input :t:`value`
+at the site of a :t:`call expression` or a :t:`method call expression` to
+a pattern.
```
kirtchev-adacore, line 3437: This might need a glossary entry. I don't see a definition of `closure parameter` in Expressions.

src/functions.rst
```diff
@@ -86,7 +84,8 @@ at the site of a :t:`call expression` or a :t:`method call expression` to
 a pattern.
 
 :dp:`fls_lxzinvqveuqh`
-A :t:`function parameter` is an :t:`irrefutable pattern`.
+The :t:`pattern` of a :t:`function parameter` if it exists shall be
```
kirtchev-adacore, line 87: I don't think that "if it exists" is necessary. If there is no function parameter, then the rule does not apply.

kirtchev-adacore (changes requested):

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #257, opened 2022-08-17
Define undefined behavior for valid type values in terms of invariants

Veykril:
We currently specify non-utf8 str values as UB which is not quite the case,
this PR changes how we describe undefined behavior for valid values of a type
by specifying either a validity invariant or a safety invariant for them.

Closes https://github.com/ferrocene/specification/issues/150

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -4810,9 +4810,8 @@ safety invariant
 ^^^^^^^^^^^^^^^^
 
 :dp:`fls_wRZfAmTmMGTX`
-A :dt:`safety invariant` is an invariant that programs are free to assume to be
-upheld for their execution and when violated may result in :t:`undefined
-behavior`.
+A :dt:`safety invariant` is an invariant that when violated may result in
+:t:`undefined behavior`.
 
 .. _fls_XeMNghZZOBqL:
 
@@ -5935,7 +5934,7 @@ undefined behavior
 ^^^^^^^^^^^^^^^^^^
 
 :dp:`fls_WpwmltUMQGZa`
-Situations that result in unbounded errors.
+:dt:`Undefined behavior`` is a situation that results in an unbounded error.
 
 .. _fls_57kis2vnt3cv:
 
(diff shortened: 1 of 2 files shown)


src/glossary.rst
```diff
@@ -4804,6 +4804,16 @@ rustc
...
 
+.. _fls_Q4MRIo7cWv5K:
+
+safety invariant
+^^^^^^^^^^^^^^^^
+
+:dp:`fls_wRZfAmTmMGTX`
+A :dt:`safety invariant` is an invariant that programs are free to assume to be
+upheld for their execution and when violated may result in :t:`undefined
+behavior`.
```
kirtchev-adacore, line 4815: How about something along the lines of

A safety invariant is an invariant that when violated may result in undefined behavior.

src/glossary.rst
```diff
@@ -5919,6 +5929,14 @@ unary operator
 :dp:`fls_p6mk2zrwgwem`
 A :dt:`unary operator` operates on one :t:`operand`.
 
+.. _fls_WuLL4SvSKavZ:
+
+undefined behavior
+^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_WpwmltUMQGZa`
+Situations that result in unbounded errors.
```
kirtchev-adacore, line 5938: This would need a `:dt:`. How about

Undefined behavior is a situation that results in an unbounded error.

src/glossary.rst
```diff
@@ -6301,6 +6319,15 @@ valid
...
 reachable control flow paths.
 
+.. _fls_A5K8aOBsI3BG:
+
+validity invariant
+^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_3ebC3l839ajF`
+A :dt:`validity invariant` is an invariant that when not uphold will result in
+immediate :t:`undefined behavior`.
```
kirtchev-adacore, line 6329: How about

A validity invariant is an invariant that when violated results in immediate undefined behavior.

kirtchev-adacore (changes requested):
I like this invariant business!

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #258, opened 2022-08-17
Add lint to ensure the glossary is ordered

pietroalbini:
This PR adds an alternate implementation of #249 by creating a `ferrocene_spec_lints` extension walking the AST, rather than ad-hoc parsing the glossary file. This should be more resilient, and doesn't hardcode the structure of `glossary`. We'll be able to use the same extension for future lints we'll want to implement (like ensuring all paragraphs and sections have an ID).

The second and third commit also improve `make.py` to fail the build if warnings are present and to hide the useless backtrace when `sphinx-build` fails. The last commit is cherry-picked from #249 and fixes the glossary ordering errors.


Veykril (approved):

Veykril:
bors merge

PR #264, opened 2022-08-19
Correct Send conditions for closure captures

tmandry:
If a closure's by-move captures are Send but not Sync, the closure can still be Send.

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a2d1fa6212d2ab1b486fe5823fcc70a7)

```rust
fn expect_send(_: &impl Send) {}
fn expect_sync(_: &impl Sync) {}

struct Foo(*mut ());
unsafe impl Send for Foo {}

fn main() {
    let foo = Foo(0x0 as _);
    expect_send(&foo);
    // Fails:
    // expect_sync(&foo);

    let cl = || {
        let _x = foo; // replace by &foo to fail
    };
    expect_send(&cl);
    // Fails:
    // expect_sync(&cl);
}
```

### DIFF
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -869,8 +869,8 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Send`
 
 * :dp:`fls_vamgwed199ct`
   The :t:`[type]s` of all :t:`[capture target]s` that employ :t:`by immutable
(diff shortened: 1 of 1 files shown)


Veykril:
Thanks!
bors merge

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #266, opened 2022-08-26
Mark informational pages in the source code

pietroalbini:
This PR adds a new directive, `informational-page`, which marks the page as informational. An informational page:

* ...will have a blue banner at the top explaining the page is informational.
* ...will be marked as such in `paragraph-ids.json`, so that the rest of our internal tooling can understand the page doesn't contain normative paragraphs.

image


pietroalbini:
bors try

kirtchev-adacore (approved):
Excellent, exactly what we needed!

kirtchev-adacore:
bors merge

PR #269, opened 2022-08-26
Describe the CRLF to LF conversion behavior

Veykril:
Closes https://github.com/ferrocene/specification/issues/172

Does not mention doc comments, as that is part of the desugaring which I am writing as a follow up


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #271, opened 2022-08-26
dyn prefix is not optional in 2021 edition

Veykril:
Since we are going to target the 2021 edition specifically
Closes https://github.com/ferrocene/specification/issues/151


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #272, opened 2022-08-26
Clarify panic behavior

Veykril:
Closes https://github.com/ferrocene/specification/issues/171

Clarifies that execution continues on the current thread. We also do not need to explain what the function might do, it can do whatever the user wants that the language allows them to.


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #273, opened 2022-08-29
Define structural equality for patterns

Veykril:


src/glossary.rst
```diff
@@ -5357,6 +5357,15 @@ struct value
...
 A :dt:`struct value` is a :t:`value` of a :t:`struct type`.
 
+.. _fls_P7920ALJisrH:
+
+structurally equal
+^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_glRZUKhmaWmP`
+A :dt:`structurally equal` :t:`type` can be :t:`pattern matched <pattern
+matching>` on via the use of an :t:`associated constant` or :t:`constant`.
```
kirtchev-adacore, line 5367: This doesn't read right. It sounds like "structural equality" is a property of types, where one of the effects is the ability to match against a [n associated] constant (I may have misunderstood this).

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,39 @@ obtained as follows:
...
+~~~~~~~~~~~~~~~~~~~
+
+.. rubric:: Legality Rules
+
+:dp:`fls_uVTpA7gbLCYX`
+A :t:`structurally equal` :t:`type` can be :t:`pattern matched <pattern
+matching>` on via the use of an :t:`associated constant` or :t:`constant`.
+
+:dp:`fls_emcNJzl2tHSA`
+:t:`[Bool]`, :t:`Char`, :t:`[integer type]s`, :t:`str`, :t:`[function pointer
```
kirtchev-adacore, line 2011: The Bool, char, and str references should probably use :c: to get the code-like formatting.

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,39 @@ obtained as follows:
...
+type]s` and :t:`[raw pointer type]s` are :t:`structurally equal`.
+
+:dp:`fls_HpWSAfaTA1Dz`
+An :t:`abstract data type` is :t:`structurally equal` if it implements the
+:t:`[trait]s` :std:`core::marker::PartialEq` and `core::marker::Eq` via the use
```
kirtchev-adacore, line 2016: The Eq path is missing :std:. And lexicographical order :D

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,39 @@ obtained as follows:
...
+of the :t:`derive` attribute.
```
kirtchev-adacore, line 2017: of attribute :t:`derive`.

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,39 @@ obtained as follows:
...
+
+:dp:`fls_5RcnETrW6f9m`
+:t:`[Slice type]s` and :t:`[array type]s` are :t:`structurally equal` if the
```
kirtchev-adacore, line 2020: (lexicographical police): Array types and slice types

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,39 @@ obtained as follows:
...
+:t:[`element type]` contained within is :t:`structurally equal`.
```
kirtchev-adacore, line 2021: I would rephrase this as

An array type or a spice type is structurally equal if its element type is structurally equal.

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,39 @@ obtained as follows:
...
+
+:dp:`fls_hBFlaUrrhqZ6`
+:t:`[Tuple type]s` are :t:`structurally equal` if the inner :t:`[type]s` are
```
kirtchev-adacore, line 2024: By "inner type", did you mean the types of the tuple fields, or the types of the tuple fields and the types of their fields and so on, recursively?

src/glossary.rst
```diff
@@ -5357,6 +5357,15 @@ struct value
 :dp:`fls_YmZfW9kWlbIX`
 A :dt:`struct value` is a :t:`value` of a :t:`struct type`.
 
+.. _fls_P7920ALJisrH:
+
+structurally equal
```
kirtchev-adacore, line 5362: After reading all the changes, I think we should discuss the term "structurally equal" and its semantics.

kirtchev-adacore (changes requested):

src/glossary.rst
```diff
@@ -5357,6 +5357,15 @@ struct value
...
+^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_glRZUKhmaWmP`
+A :dt:`type` is a :t:`structurally equal` when its :t:`values` can be compared
```
kirtchev-adacore, line 5366: The "a" requires a noun - a structurally equal what? Did you mean "structurally equal type", or "a type has structural equality when ..."?

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,41 @@ obtained as follows:
...
+
+.. _fls_emcNJzl2tHSA:
+
+Structural Equality
+~~~~~~~~~~~~~~~~~~~
+
+.. rubric:: Legality Rules
+
+:dp:`fls_uVTpA7gbLCYX`
+A :t:`type` is a :t:`structurally equal` when its :t:`values` can be compared
```
kirtchev-adacore, line 2007: Same issue with the "a" here.

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,41 @@ obtained as follows:
...
+for equality by structure.
+
+:dp:`fls_2DZAP6JJjJ9h`
+The following :t:`[type]s` are structurally equal:
```
kirtchev-adacore, line 2011: Structurally equal should be a reference to the term.

src/types-and-traits.rst
```diff
@@ -1995,6 +1995,41 @@ obtained as follows:
...
+
+* :dp:`fls_emcNJzl2tHSA`
+  :c:`Bool`, :c:`char`, :t:`[function pointer type]s`, :t:`[integer type]s`,
+  :c:`str`,  and :t:`[raw pointer type]s`.
+
+* :dp:`fls_HpWSAfaTA1Dz`
+  An :t:`abstract data type` if it implements the :t:`[trait]s`
+  :std:`core::cmp::Eq` and :std:`core::cmp::PartialEq` via the use of the
+  :t:`[derive macro]s` :std:`core::cmp::Eq` and :std:`core::cmp::PartialEq`.
```
kirtchev-adacore, line 2020: Oh, it is no longer needed to use attribute derive?

kirtchev-adacore (changes requested):

src/types-and-traits.rst
Veykril: Well, you need to use it to invoke the `derive macros`. But my previous phrasing was incorrect as that would imply you could write a custom `derive macro` that implements them which is not the case, only the tool builtin derives have the desired effect

kirtchev-adacore: Gotcha.

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #276, opened 2022-08-30
Relax restrictions on #[no_mangle] targets

Veykril:
It is not a hard error to use the attribute on things other than statics
and functions, it just doesn't do anything for those.


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #277, opened 2022-08-30
Fix outstanding doc attribute issues

Veykril:
As it turns out, the attribute currently accepts any syntax when used as its
`doc(...)` form, and that the rejection is a future compat lint so we can just
discard this and accept anything with our current stance regarding these lints.

Closes https://github.com/ferrocene/specification/issues/117
Closes https://github.com/ferrocene/specification/issues/118

### DIFF
--- a/src/attributes.rst
+++ b/src/attributes.rst
@@ -34,7 +34,7 @@ Attributes
        SimplePath AttributeInput?
 
    AttributeInput ::=
-       DelimitedTokenTree
+       $$($$ TokenTree* $$)$$
      | $$=$$ Expression
 
    AttributeContentList ::=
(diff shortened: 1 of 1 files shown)


src/attributes.rst
```diff
@@ -866,61 +866,24 @@ Attribute ``doc``
 .. syntax::
 
    DocContent ::=
```
kirtchev-adacore, line 868: One possibility might be

DocContent ::=
    $$doc$$ DocInput

DocInput ::=
    $$=$$ MacroInvocation
  | $$=$$ StringLiteral
  | $$($$ TokenTree* $$)$$

kirtchev-adacore (changes requested):

kirtchev-adacore:
bors merge

PR #280, opened 2022-09-05
Stop tracking references in the environment

pietroalbini:
Before this commit we were tracking all references in the environment, but that is causing a strange bug in the Ferrocene monorepo when sometimes the amout of references stored in the enviornment reach 1.4 million, slowing Sphinx to a crawl.

I couldn't identify why that happens, but we're not actually using those references stored in the environment anywhere. Because of that, this commit removes tracking references in the environment.


Veykril (approved):

Veykril:
bors merge

PR #283, opened 2022-09-07
fix typo: `validty invariant` should be `validity invariant`

ZhangHanDong:
[# issue 282](https://github.com/ferrocene/specification/issues/282)

Here `validty invariant` should be a `safety invariant`, right? because it may cause UB

### DIFF
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -48,6 +48,8 @@ those :t:`[value]s`.
 Type Classification
 -------------------
 
+.. informational-section::
+
 .. rubric:: Legality Rules
 
 :dp:`fls_c4xe3pkn0n3o`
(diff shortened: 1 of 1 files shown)


Veykril:
`validity invariant` here is correct. As per our definition violated `safety invariant`s *may* cause UB at some point in the programs execution, while violated `validity invariant`s *are* immediate UB.
A function pointer in rust is required to never be null, so this is a validity invariant. A `safety invariant` example would be `str` containing utf-8 valid bytes. This can be violated, but libraries are allowed to expect valid utf-8 in `str` which may cause those libraries to invoke undefined behavior.

Though there is a typo in the documentation here still.

ZhangHanDong:
@Veykril Thanks for the explanation, but don't close this pr for now, I'll look for ralfj's article and study it before making a final confirmation

ZhangHanDong:
@Veykril  you're right, Thanks. Then this PR will be changed to fix typo.

Veykril:
Thanks for double checking on this 👍

Veykril (approved):

Veykril:
bors merge

riking:
Another, perhaps more illustrative, example of a safety invariant is that a function pointer points to a valid function that is possible to call.

PR #284, opened 2022-09-07
Describe when temporaries are created in expressions

Veykril:
This is only half of the picture, pattern matching can also introduce temporaries, but the rules there are more complex to describe


Veykril:
I am also not quite sure about the place, as said, patterns can also introduce temporaries, so should this go into the values chapter on temporaries? Or should we explain creation for expressions and patterns separately like this is going?

kirtchev-adacore (approved):
Let us keep the new rule in the Value Expressions section.

kirtchev-adacore:
bors merge

PR #291, opened 2022-09-08
Resolve broken links and other minor issues in Chapter 7.

kirtchev-adacore:


src/values.rst
```diff
@@ -130,11 +130,12 @@ A :t:`static` is a :t:`value` that is associated with a specific memory
...
 :dp:`fls_mt94jvoot9dx`
-A :t:`static` defined within a :t:`generic` is declared once and shared between
-all :t:`[instantiation]s`.
+A :t:`static` defined within a :t:`generic function` exists once in the
+output executable or library.
 
 :dp:`fls_k0r2c6uq29tu`
-The :t:`type specification` of a :t:`static` shall have a :t:`static lifetime`.
+The :t:`type specification` of a :t:`static` shall have a ``'static``
+:t:`lifetime`.
```
Veykril, line 138:
```suggestion
The :t:`type specification` of a :t:`static` shall have ``'static``
:t:`lifetime`.
```
"a 'static lifetime` implies to me that it needs to specify `'static` in it somewhere, which isn't really the case. What this is about is that the type should be alive/valid for the entire duration of the program

kirtchev-adacore: Good point.

Veykril (approved):

Veykril:
bors merge

PR #300, opened 2022-09-12
Specifiy implied lifetime bounds

Veykril:
Most likely not quite correct, but the exact rules around this are rather complex and thats the best I could make out that should generally hold.
https://github.com/rust-lang/rfcs/blob/master/text/1214-projections-lifetimes-and-wf.md
https://github.com/rust-lang/rfcs/blob/master/text/0192-bounds-on-object-and-generic-types.md
https://github.com/rust-lang/rfcs/blob/master/text/2093-infer-outlives.md

### DIFF
--- a/src/generics.rst
+++ b/src/generics.rst
@@ -177,34 +177,36 @@ A :t:`type parameter` has an implicit :std:`core::marker::Sized` :t:`bound`, unl
 ``?core::marker::Sized`` bound is present.
 
 :dp:`fls_yaYedRNaxhTe`
-A :t:`type parameter` declared on an :t:`abstract data type` has implicit
-:t:`lifetime` :t:`[bound]s` depending on its usages in the :t:`abstract data
-type`'s fields as follows:
+A :t:`type parameter` of an :t:`abstract data type` has implicit
+:t:`lifetime` :t:`[bound]s` depending on its usage in the :t:`[field]s` of the
+:t:`abstract data type` as follows:
 
 * :dp:`fls_axMVssU7an8p`
-  If the :t:`type parameter` is used in a :t:`reference type` it has an implicit
-  `lifetime` :t:`bound` for the :t:`reference type`'s lifetime,
+  If the :t:`type parameter` is used in a :t:`reference type`, then the :t:`type
+  parameter` has an implicit `lifetime` :t:`bound` for the :t:`lifetime` of the
+  :t:`reference type`,
 
 * :dp:`fls_iAnCoxISuxxb`
   If the :t:`type parameter` is used as a :t:`generic argument` of an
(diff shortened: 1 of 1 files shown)


src/generics.rst
```diff
@@ -173,9 +173,39 @@ A :t:`constant parameter` shall be used in the following contexts:
   As a :t:`value` of any :t:`expression` within a :t:`function body`.
 
 :dp:`fls_hidfwkwr2r73`
-A :t:`type parameter` has an implicit :std:`core::marker::Sized` bound, unless a
+A :t:`type parameter` has an implicit :std:`core::marker::Sized` :t:`bound`, unless a
 ``?core::marker::Sized`` bound is present.
 
+:dp:`fls_yaYedRNaxhTe`
+A :t:`type parameter` declared on an :t:`abstract data type` has implicit
```
kirtchev-adacore, line 180: How do you like

A type parameter of an abstract data type ...

I find "declared on" a bit weird.

src/generics.rst
```diff
@@ -173,9 +173,39 @@ A :t:`constant parameter` shall be used in the following contexts:
...
+:t:`lifetime` :t:`[bound]s` depending on its usages in the :t:`abstract data
+type`'s fields as follows:
```
kirtchev-adacore, line 182: in the fields of the abstract data type, as follows:

src/generics.rst
```diff
@@ -173,9 +173,39 @@ A :t:`constant parameter` shall be used in the following contexts:
...
+
+* :dp:`fls_axMVssU7an8p`
+  If the :t:`type parameter` is used in a :t:`reference type` it has an implicit
```
kirtchev-adacore, line 185: How about

If the type parameter is used in a reference type, then the type parameter has an implicit lifetime bound for the lifetime of the reference type.

src/generics.rst
```diff
@@ -173,9 +173,39 @@ A :t:`constant parameter` shall be used in the following contexts:
...
+  `lifetime` :t:`bound` for the :t:`reference type`'s lifetime,
+
+* :dp:`fls_iAnCoxISuxxb`
+  If the :t:`type parameter` is used as a :t:`generic argument` of an
+  :t:`abstract data type`, the :t:`type parameter` inherits all the
+  :t:`lifetime` :t:`[bound]s` of the corresponding :t:`generic parameter` of the
```
kirtchev-adacore, line 191: inherits all lifetime bounds **from** the corresponding ...

kirtchev-adacore (changes requested):

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

Issue #296, opened 2022-09-09
make.py fails with DefinitionCollector exception


kirtchev-adacore:
The following [commit](https://github.com/ferrocene/specification/pull/295/commits/09546d9f1283478bda0696b26674e60598d90bda) causes `/make.py -c` to fail with
```
reading sources... [ 95%] unsafety                                              
Sphinx parallel build error:
sphinx.errors.ExtensionError: Handler <bound method DefinitionsCollector.process_doc of <ferrocene_spec.definitions.DefinitionsCollector object at 0x7fd75462ae20>> for event 'doctree-read' threw an exception (exception: )
```
It would be nice if the exception hinted at what the problem might be.

Veykril:
Additional info, this exception gets thrown when an ID for a section header is missing the leading underscore.

PR #302, opened 2022-09-13
Handle sections without an ID

pietroalbini:
Reported in #296, if someone puts a paragraph with an ID inside a section without an ID, an exception happens because the extension can't link the paragraph with its parent section. This PR adds a warning explaining what the problem is and where it happen. Note that even with this patch, an exception will be raised down the line, but there will at least be a warning showing where the error comes from.

This also adds a `--debug` flag to `make.py` that shows the exception. Showing exceptions is behind a flag because showing exception with parallel builds enabled will result in an useless exception being reported due to how Sphinx's multiprocessing works. So, the `--debug` flag also disables parallel builds, which slows down things a lot.

### DIFF
--- a/exts/ferrocene_spec/definitions/paragraphs.py
+++ b/exts/ferrocene_spec/definitions/paragraphs.py
@@ -5,6 +5,7 @@ from .. import utils
 from collections import defaultdict
 from docutils import nodes
 import hashlib
+import sphinx
 
 
 ROLE = "p"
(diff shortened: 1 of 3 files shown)


exts/ferrocene_spec/definitions/paragraphs.py
```diff
@@ -49,7 +50,16 @@ def collect_items_in_document(app, nodes_to_collect):
         if section_node is None:
             raise RuntimeError(f"could not find section for {node!r}")
 
-        section_id, section_anchor = utils.section_id_and_anchor(section_node)
+        try:
+            section_id, section_anchor = utils.section_id_and_anchor(section_node)
+        except utils.NoSectionIdError:
+            logger = sphinx.util.logging.getLogger(__name__)
+            logger.warn(
+                "paragraph inside a section tha doesn't have an id starting with fls_",
```
Veykril, line 58:
```suggestion
                "paragraph inside a section that doesn't have an id starting with fls_",
```

Veykril (dismissed):

Veykril (approved):

Veykril:
bors merge

PR #304, opened 2022-09-13
Explain conformance between [method] call expression and function

Veykril:
Closes https://github.com/ferrocene/specification/issues/279

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -3052,10 +3052,14 @@ a left-to-right order.
    let u1 = Union { int: 0 };
    let u2 = Union { float: 0.0 };
 
+
+Call Expressions and Method Call Expressions
+--------------------------------------------
+
 .. _fls_xa4nbfas01cj:
 
 Call Expressions
-----------------
+~~~~~~~~~~~~~~~~
 
 .. rubric:: Syntax
 
@@ -3200,7 +3204,7 @@ It is undefined behavior to call a :t:`function` with an :t:`ABI` other than the
 .. _fls_z7q8kbjwdc7g:
 
 Method Call Expressions
------------------------
+~~~~~~~~~~~~~~~~~~~~~~~
 
 .. rubric:: Syntax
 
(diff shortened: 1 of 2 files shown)


src/expressions.rst
```diff
@@ -3100,6 +3100,20 @@ a :t:`tuple enum variant` or a :t:`type` that implements any of the
...
 :t:`callee type` is found, otherwise this is a static error.
 
+:dp:`fls_c40C6rg6rGv6`
+An :t:`argument operand` of a :t:`call expression` matches a :t:`parameter` or
+:t:`field` of the :t:`callee type` when its position and the position of the
+:t:`parameter` or :t:`field` of the :t:`callee type` are the same. Such an
+:t:`argument operand` is a :dt:`matched argument operand`.
+
+:dp:`fls_Gr1ixJ9vFjUm`
+The :t:`type` of an :t:`argument operand` of a :t:`call expression` and the
```
kirtchev-adacore, line 3110: You probably meant "matched argument operand" here. Here is an alternative wording:

The type of a matched argument operand and the type of the corresponding function parameter or field shall be unifiable.

kirtchev-adacore (changes requested):
Here is an alternative strategy to these rules:

1. Create a common section "Call Expressions and Method Call Expressions".
2. Make the existing sections on call expressions and method call expressions be subsections of the new common section.
3. Add section "Call Conformance" to the new section.
4. Somewhere in Call Conformance, indicate that a method call expression is equivalent to a call expression where the receiver is the first argument (or prepended to the argument list). This will eliminate the whole "offset by 1" business.
5. From then on, you will just need to explain call conformance.

src/expressions.rst
```diff
@@ -3052,10 +3052,14 @@ a left-to-right order.
    let u1 = Union { int: 0 };
    let u2 = Union { float: 0.0 };
 
+
```
kirtchev-adacore, line 3055: This section should have a `.. _fls_xxx:` ID.

src/expressions.rst
```diff
@@ -3266,6 +3270,29 @@ The :t:`evaluation` of a :t:`method call expression` proceeds as follows:
...
 
+.. _fls_8gPCPVc99pXJ:
+
+Call Conformance
+~~~~~~~~~~~~~~~~
+
+:dp:`fls_tsn6SUUG9LvW`
+A :t:`method call expression` is equivalent to a :t:`call expression` where the
+:t:`call operand` is the resolved :t:`method` and the adjusted :t:`receiver
+operand` prepended to the :t:`[argument operand]s`.
```
kirtchev-adacore, line 3281: ... the adjusted receiver operand **is** prepended to ...

src/expressions.rst
```diff
@@ -3266,6 +3270,29 @@ The :t:`evaluation` of a :t:`method call expression` proceeds as follows:
...
+
+:dp:`fls_c40C6rg6rGv6`
+An :t:`argument operand` matches a :t:`function parameter` or :t:`field` of the :t:`callee type` when its position and the position of the
+:t:`parameter` or :t:`field` of the :t:`callee type` are the same. Such an
```
kirtchev-adacore, line 3285: No need to repeat "of the callee type" because we already identified which function parameter and field we are talking about.

Note that :t:`parameter` should be :t:`function parameter`.

src/glossary.rst
```diff
@@ -937,6 +937,14 @@ C representation
 :dt:`C representation` is a :t:`type representation` that lays out :t:`[type]s`
 such that they are interoperable with the :t:`C` language.
 
+.. _fls_Egfa8tdbqllA:
+
+Call conformance
+^^^^^^^^^^^^^^^^
+:dp:`fls_Jr1gUX7Ju4Oh`
```
kirtchev-adacore, line 944: Add an empty new line preceding this one.

kirtchev-adacore (changes requested):

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #310, opened 2022-09-16
Clarify the site of fragment specifier restrictions

kirtchev-adacore:


kirtchev-adacore:
@Veykril
As currently stated, the fragment specifier restrictions impose a set of legal tokens that must follow a metavariable at the macro invocation site.
Note that the augmented example does not need to use `=>`, `,`, or `;` in the macro invocation when the fragment specifier is `expr`. Am I missing something here?

src/macros.rst
```diff
@@ -151,7 +151,8 @@ a :t:`metavariable`.
 
 :dp:`fls_8o9mcV2KrKac`
 :t:`Fragment specifier` kinds impose the following
-:dt:`[fragment specifier restruction]s` on the :t:`[token]s` that follow them:
+:dt:`[fragment specifier restriction]s` on the :t:`[token]s` that follow them
+at the :t:`macro invocation` site:
```
Veykril, line 155: That's not correct, these restrictions are checked at the macro declaration site, https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=42e223f70885d65d3a3e8b63b980b730

PR #311, opened 2022-09-16
Fix example stating static when it should be type alias

Veykril:
https://github.com/ferrocene/specification/pull/301/files#r972485521


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #315, opened 2022-09-22
Define scoping visibility metavariables

Veykril:


src/macros.rst
```diff
@@ -145,6 +145,10 @@ transcription`.
 :dp:`fls_4zdait30exvn`
 A :t:`metavariable` is a :t:`macro match` that describes a :t:`variable`.
 
+:dp:`fls_2HguXbL7DjKH`
+A :t:`metavariable` is visible in the :t:`macro transcriber` of the :t:`macro
+rule` of the :t:`macro matcher` it is declared in.
```
Veykril, line 150: Do we used `declare` or `define` for these things, I am actually not sure if we even use one uniformly

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #316, opened 2022-09-22
Specify binding and variable without mutually referring to each other

Veykril:
We currently define them by using each other, which is obviously a bad definition.

### DIFF
--- a/src/values.rst
+++ b/src/values.rst
@@ -226,10 +226,7 @@ The following :t:`[construct]s` are :t:`[variable]s`:
   A :t:`temporary`.
 
 * :dp:`fls_81dlbula47nu`
-  A named :t:`binding`.
-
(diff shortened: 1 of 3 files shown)


src/values.rst
```diff
@@ -226,10 +226,7 @@ The following :t:`[construct]s` are :t:`[variable]s`:
   A :t:`temporary`.
 
 * :dp:`fls_81dlbula47nu`
-  A named :t:`binding`.
-
-* :dp:`fls_adqfhc5k051x`
-  A named :t:`function parameter`.
```
Veykril, line 232: A function parameter introduces bindings which are the variables, I wouldn't consider the parameters themselves as variables hence

kirtchev-adacore:
bors merge

PR #320, opened 2022-09-26
Mention field resolution in field access expressions

Veykril:
Also removes the subsections, as the indexed one has no purpose anymore (the validity of the index should be encoded in the field  resolution already)


kirtchev-adacore (approved):
Note that the tests will need to be updated as they refer to the Named and Indexed Field Access sections.

kirtchev-adacore:
bors merge

PR #325, opened 2022-09-30
References to unaligned fields are an error

jonas-schievink:
Fixes https://github.com/ferrocene/specification/issues/144


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

Veykril:
@kirtchev-adacore I believe we said we would allow future compat lints?

kirtchev-adacore:
@Veykril The consensus is that we will describe the behavior of the baseline (1.66 or 1.67), whatever that behavior might be. If something is flagged with a warning at the time of cutting the baseline and it is not yet an error, then it is technically legal.

Veykril:
Ah I didn't realize this lint changed to error by default by now

PR #326, opened 2022-09-30
Fix reference to `ops::Sub` trait

jonas-schievink:


kirtchev-adacore:
Good catch!

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #327, opened 2022-09-30
Remove the term "value holder"

jonas-schievink:
Part of https://github.com/ferrocene/specification/issues/178 (not yet sure what to do about "owner")

Since only local variables participate in initialization tracking, and ownership, but the term *value holder* includes consts and statics, the term is incorrect here.

This is still inaccurate, since borrows and initialization is tracked on a per-field basis in some cases, but already an improvement.

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -4411,7 +4411,7 @@ owner
 ^^^^^
 
 :dp:`fls_7vwwhberexeb`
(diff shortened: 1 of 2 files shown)


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #328, opened 2022-09-30
Define and list all `auto trait`s and specify coinductive matching

jonas-schievink:
...to the best of my ability

Fixes https://github.com/ferrocene/specification/issues/122

Part of this arguably belongs in the *Implementations* chapter, but I kept it in the *Traits* subchapter so that it's closer together.


Veykril (approved):

Veykril:
bors merge

PR #330, opened 2022-10-04
Specify built-in arithmetic rules and overflow

jonas-schievink:
Fixes https://github.com/ferrocene/specification/issues/201

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -1281,8 +1281,20 @@ The :t:`evaluation` of an :t:`addition expression` proceeds as follows:
 #. :dp:`fls_jjmc1xgny77`
    The :t:`right operand` is evaluated.
 
+#. :dp:`fls_NcLf4o1dpniS`
+   If the types of both operands are the same :t:`integer type` or
+   :t:`floating-point type`, the :t:`addition expression` evaluates to the sum
+   of the operands, following the rules of unsigned integer addition for
+   :t:`[unsigned integer type]s`, two's complement addition for
+   :t:`[signed integer type]s`, or floating-point addition for
+   :t:`[floating-point type]s`.
+
+   #. :dp:`fls_zeGcAQBRcTJs`
+      If unsigned integer addition or two's complement addition is performed,
+      the operation may result in an :t:`arithmetic overflow`.
+
 #. :dp:`fls_cayhj5hcuhcg`
-   ``core::ops::Add::add(left_operand, right_operand)`` is invoked.
+   Otherwise, ``core::ops::Add::add(left_operand, right_operand)`` is invoked.
 
 :dp:`fls_43knkymqpj7t`
 The :t:`evaluation` of a :t:`division expression` proceeds as follows:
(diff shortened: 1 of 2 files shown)


src/expressions.rst
```diff
@@ -1281,8 +1281,20 @@ The :t:`evaluation` of an :t:`addition expression` proceeds as follows:
 #. :dp:`fls_jjmc1xgny77`
    The :t:`right operand` is evaluated.
 
+#. :dp:`fls_NcLf4o1dpniS`
+   If the types of both operands are the same :t:`integer type` or
```
kirtchev-adacore, line 1285: I would change the beginning of the sentence to

If the :t:`type` of both :t:`[operand]s` is the same ...

src/expressions.rst
```diff
@@ -1281,8 +1281,20 @@ The :t:`evaluation` of an :t:`addition expression` proceeds as follows:
...
+   :t:`floating-point type`, the :t:`addition expression` evaluates to the sum
+   of the operands, following the rules of unsigned integer addition for
```
kirtchev-adacore, line 1287: :t:`[operand]s`

src/expressions.rst
```diff
@@ -1293,8 +1305,25 @@ The :t:`evaluation` of a :t:`division expression` proceeds as follows:
 #. :dp:`fls_bveocgaagk1n`
    The :t:`right operand` is evaluated.
 
+#. :dp:`fls_zLroZh43MOtN`
+   If the types of both operands are the same :t:`integer type` or
```
kirtchev-adacore, line 1309: Similar issues as the paragraphs above.

src/expressions.rst
```diff
@@ -4813,3 +4884,25 @@ within the :t:`capturing expression`, as follows:
...
 A tool selects the first :t:`capture mode` that is compatible with the use of
 the :t:`capture target`.
+
+.. _fls_ZfIBiJMf8qE1:
+
+Arithmetic Overflow
+-------------------
+
+:dp:`fls_oFIRXBPXu6Zv`
+An :dt:`arithmetic overflow` occurs if an :t:`arithmetic expression` or a
```
kirtchev-adacore, line 4894: "Arithmetic overflow" should be defined in the Glossary. The sentence should also be left here as an introduction, but the reference to "arithmetic overflow" should be changed to :t:.

src/expressions.rst
```diff
@@ -4813,3 +4884,25 @@ within the :t:`capturing expression`, as follows:
...
+:t:`negation expression` computes a :t:`value` of a :t:`scalar type` that lies
+outside of the range of valid :t:`[value]s` for the :t:`scalar type`.
+
+.. rubric:: Dynamic Semantics
+
+:dp:`fls_8fU3h8PLasqA`
+There are two allowed behaviors for :t:`arithmetic overflow`:
+
+#. :dp:`fls_R48VKcEIbfXC`
+   Evaluation of the expression may result in a :t:`panic`.
```
kirtchev-adacore, line 4904: :t:`Evaluation` of the :t:`expression`

src/expressions.rst
```diff
@@ -4813,3 +4884,25 @@ within the :t:`capturing expression`, as follows:
...
+
+#. :dp:`fls_QMpI8K43K2yU`
+   The resulting value of the expression may be truncated, discarding the most
```
kirtchev-adacore, line 4907: The resulting :t:`value` of the :t:`expression` may be truncated, discarding the most significant bits that do not fit in the target :t:`type`.

kirtchev-adacore (changes requested):

src/expressions.rst
jonas-schievink: Is there any way to link from the glossary to this paragraph? That might be more convenient for finding it

kirtchev-adacore: Unfortunately no. I raised this issue in the appropriate Zulip channel. Whatever the solution, it would encompass the whole Glossary. In the mean time, you should still define `arithmetic overflow` in the Glossary.

jonas-schievink: Okay, did that

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #333, opened 2022-10-06
Specify opt-out trait bounds

Veykril:
Closes https://github.com/ferrocene/specification/issues/77

### DIFF
--- a/src/generics.rst
+++ b/src/generics.rst
@@ -48,20 +48,12 @@ Generic Parameters
 
 :dp:`fls_sye3d17l9bf5`
 A :t:`generic parameter` is a placeholder for a :t:`constant`, a :t:`lifetime`,
-or a :t:`type` whose :t:`value` is supplied statically by a
-:t:`generic argument`.
-
-:dp:`fls_4a2qshaf5se7`
-It is a static error to use a :t:`generic parameter` in the
-:t:`discriminant initializer` of an :t:`enum variant`.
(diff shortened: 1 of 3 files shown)


src/generics.rst
```diff
@@ -48,21 +48,13 @@ Generic Parameters
 
 :dp:`fls_sye3d17l9bf5`
 A :t:`generic parameter` is a placeholder for a :t:`constant`, a :t:`lifetime`,
-or a :t:`type` whose :t:`value` is supplied statically by a
-:t:`generic argument`.
-
-:dp:`fls_4a2qshaf5se7`
-It is a static error to use a :t:`generic parameter` in the
-:t:`discriminant initializer` of an :t:`enum variant`.
+or a :t:`type` whose :t:`constant`, :t:`lifetime`, or :t:`type` is supplied
```
kirtchev-adacore, line 51: This is a bit confusing, it looks like the type has a constant, a lifetime, or a type. Why doesn't value work here?

kirtchev-adacore (changes requested):

src/generics.rst
Veykril: A value is an instance of a type. What one is supplying to a lifetime parameter is a lifetime though, for a type parameter you supply a type, and for a constant parameter you supply a constant (value). So value is not the correct term here

kirtchev-adacore: Good point! See suggestion.

kirtchev-adacore (approved):

PR #339, opened 2022-10-11
Clarify "transparent representation"

jonas-schievink:
On the ABI level, a `#[repr(transparent)]` struct is equivalent to its only non-zero-sized field, not to a struct containing that field, which is passed differently on some architectures. We don't really specify ABI outside of what `extern "abi"` means, but this should still clarify that.


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #359, opened 2022-10-18
Clean up Chapter 18: Program Structure and Compilation

kirtchev-adacore:


src/glossary.rst
```diff
@@ -3927,7 +3927,8 @@ module
 ^^^^^^
 
 :dp:`fls_ujlsg58bskl5`
-A :dt:`module` is a container for zero or more :t:`[item]s`.
+A :dt:`module` is a container for zero or more :t:`[inner attribute]s`,
+:t:`[inner doc comment]s`, and :t:`[item]s`.
```
Veykril, line 3931: I don't quite like this, a module doesn't really contain attributes and doc comments in the sense that they contain items. The attributes and doc comments are part of the module, whether they are outer or inner ones.

kirtchev-adacore: Fair enough. I restored the original wording.

Veykril:
bors merge

PR #366, opened 2022-10-21
Remove "mutual supertype" wording from LUB coercions

jonas-schievink:
This should fully specify how least-upper-bound (LUB) coercions work. Notably they seem to handle function pointer types very robustly, but no other coercions like from `&[T; 0]` and `&[T; 1]` to `&[T]`.

Closes https://github.com/ferrocene/specification/issues/67 (avoids the wording entirely)


kirtchev-adacore:
bors merge

PR #367, opened 2022-10-25
Specify unification behavior when type variables are involved

jonas-schievink:
The propagation of type information across type variables during unification is what drives type inference forward, so it is quite important to at least approximate that behavior here.

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -6575,7 +6575,10 @@ type unification
 ^^^^^^^^^^^^^^^^
 
 :dp:`fls_3vyodut341b5`
-:dt:`Type unification` is a measure of compatibility between two :t:`[type]s`.
+:dt:`Type unification` is the process by which :t:`type inference` propagates
+known :t:`[type]s` across the :t:`type inference root` and assigns concrete
+:t:`[type]s` to :t:`[type variable]s`, and also a general mechanism to check
+for compatibility between two :t:`[type]s` during :t:`method resolution`.
 
 .. _fls_6zhffgxtytku:
 
@@ -6583,7 +6586,7 @@ type variable
 ^^^^^^^^^^^^^
 
 :dp:`fls_j9eusnwze4rz`
-A :t:`type variable` is a placeholder used during :t:`type inference` to stand
+A :dt:`type variable` is a placeholder used during :t:`type inference` to stand
 in for an undetermined :t:`type`.
 
 .. _fls_44uvj9l7q98z:
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -6577,7 +6577,7 @@ type unification
 :dp:`fls_3vyodut341b5`
 :dt:`Type unification` is the process by which :t:`type inference` propagates
 known :t:`[type]s` across the :t:`type inference root` and assigns concrete
-:t:`[type]s` to :t:`[type variable]s`, and also a general mechanism to check
+:t:`[type]s` to :t:`[type variable]s`, as well as a general mechanism to check
 for compatibility between two :t:`[type]s` during :t:`method resolution`.
 
 .. _fls_6zhffgxtytku:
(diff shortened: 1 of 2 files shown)


src/types-and-traits.rst
```diff
@@ -1637,13 +1637,65 @@ Type Unification
...
+A :t:`type` is said to :t:`unify` with another :t:`type` when the domains,
+ranges, and structures of both :t:`[type]s` are compatible according to the
+rules detailed below.
 
 :dp:`fls_aie0tr62vhw5`
 Two types that :t:`unify` are said to be :t:`[unifiable type]s`.
 
+:dp:`fls_3U7Ue6Xzuv9M`
+:t:`Type unification` is a symmetric operation. If a :t:`type` ``A`` unifies
+with a :t:`type` ``B``, then ``B`` also unifies with ``A`` and such
```
kirtchev-adacore, line 1655:
```suggestion
:t:`Type unification` is a symmetric operation. If :t:`type` ``A`` unifies
with :t:`type` ``B``, then ``B`` also unifies with ``A`` and such
```

src/glossary.rst
```diff
@@ -6575,15 +6575,18 @@ type unification
 ^^^^^^^^^^^^^^^^
 
 :dp:`fls_3vyodut341b5`
-:dt:`Type unification` is a measure of compatibility between two :t:`[type]s`.
+:dt:`Type unification` is the process by which :t:`type inference` propagates
+known :t:`[type]s` across the :t:`type inference root` and assigns concrete
+:t:`[type]s` to :t:`[type variable]s`, and also a general mechanism to check
```
kirtchev-adacore, line 6580:
```suggestion
:t:`[type]s` to :t:`[type variable]s`, as well as a general mechanism to check
```

src/types-and-traits.rst
```diff
@@ -1637,13 +1637,65 @@ Type Unification
 .. rubric:: Legality Rules
 
 :dp:`fls_ryvdhkgm7vzj`
-:t:`Type unification` is a measure of compatibility between two :t:`[type]s`. A
-:t:`type` is said to :t:`unify` with another :t:`type` when the domains,
-ranges, and structures of both :t:`[type]s` are compatible.
+:t:`Type unification` is the process by which :t:`type inference` propagates
+known :t:`[type]s` across the :t:`type inference root` and assigns concrete
+:t:`[type]s` to :t:`[type variable]s`, and also a general mechanism to check
```
kirtchev-adacore, line 1642:
```suggestion
:t:`[type]s` to :t:`[type variable]s`, as well as a general mechanism to check
```

src/types-and-traits.rst
```diff
@@ -1637,13 +1637,65 @@ Type Unification
...
+:dp:`fls_3U7Ue6Xzuv9M`
+:t:`Type unification` is a symmetric operation. If a :t:`type` ``A`` unifies
+with a :t:`type` ``B``, then ``B`` also unifies with ``A`` and such
+:t:`type unification` results in the same observable effects.
+
+:dp:`fls_tIiw5WkTRNf7`
+If one of the two :t:`[type]s` is a :t:`type variable`, :t:`type unification`
+proceeds as follows:
+
+* :dp:`fls_EoEbldkKBQW0`
```
kirtchev-adacore, line 1662: I suspect that these list items should be part of a numbered list as the logic implies ordering.

src/types-and-traits.rst
```diff
@@ -1637,13 +1637,65 @@ Type Unification
...
+    If ``U`` is an :t:`floating-point type` or an
+    :t:`floating-point type variable`, the :t:`floating-point type variable`
+    ``T`` is assigned :t:`type` ``U``.
+
+  * :dp:`fls_nKcqFo7yIDBe`
+    Otherwise, :t:`type unification` fails.
+
+* :dp:`fls_jkaiBnApJAtt`
+  Otherwise, neither :t:`type` is a :t:`type variable`, and the rules below
+  apply instead.
```
kirtchev-adacore, line 1697:
```suggestion
  are in effect.
```

kirtchev-adacore (changes requested):
I like the additions!

src/types-and-traits.rst
jonas-schievink: Yeah, good point. Did that.

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #379, opened 2022-11-07
Describe let else statements

### LINKED ISSUE #377: Specify let else statements

kirtchev-adacore:
Fixes #377 .

### DIFF
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -2129,26 +2129,6 @@ The following :t:`[construct]s` are considered :t:`[type inference root]s`:
 * :dp:`fls_cPlCLGCcl7EK`
   A :t:`static initializer`.
 
-:dp:`fls_uvvn4usfsbhr`
-A :t:`type variable` is a placeholder used during :t:`type inference` to stand
-in for an undetermined :t:`type`.
-
-:dp:`fls_gDalJm1XS0mi`
-A :t:`global type variable` is a :t:`type variable` that can refer to any
-:t:`type`.
-
-:dp:`fls_7ov36fpd9mwe`
-An :t:`integer type variable` is a :t:`type variable` that can refer only to
-:t:`[integer type]s`.
-
-:dp:`fls_3hv3wxkhjjp1`
-A :t:`floating-point type variable` is a :t:`type variable` that can refer only
-to :t:`[floating-point type]s`.
-
-:dp:`fls_bXQ63GYYDuMp`
-A :t:`diverging type variable` is a :t:`type variable` that can refer to any
-:t:`type` and originates from a :t:`diverging expression`.
-
 :dp:`fls_J77VeAlD8rsv`
 A :t:`type inference root` imposes an :dt:`expected type` on its :t:`expression`
 depending on the :t:`type inference root` as follows:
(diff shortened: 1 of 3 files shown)


src/types-and-traits.rst
```diff
@@ -2101,31 +2101,54 @@ Type Inference
...
-  A :t:`static initializer`.
-
 * :dp:`fls_KphY5qHev0Dc`
   The :t:`discriminant initializer` of an :t:`enum variant`.
 
 * :dp:`fls_Sowatt1V988J`
   A :t:`function body`.
 
+* :dp:`fls_dF4kA29a4iMF`
+  A :t:`let initializer`.
```
jonas-schievink, line 2124: `let` is not a type inference root, it is inferred together with all other statements in the containing root

kirtchev-adacore: I don't think so. A `let statement` has an optional `type ascription`. If the ascription is present, then the ascription imposes its type on the `let initializer` (more precisely the expression of the let initializer). Otherwise how do we trigger type inference for `let statement`s?

jonas-schievink: It's inferred as part of the containing inference root. Inference collects all type ascriptions and known types (including those in path expressions like `Option::<u8>::None`), creates inference variables for everything that isn't known yet, and then propagates the known types through inference variables via unification (and trait solving and method resolution etc.).

Ascriptions can also be incomplete, like in `let x: Option<_>`, and in that case inference has to figure out what the `_` stands for using the surrounding context (not just the initializer). Every inference root is inferred in isolation, so this would not be possible if `let` was one.

kirtchev-adacore: @Veykril Lukas, what do you think about the wording, and Jonas's points?

kirtchev-adacore: After discussing this with Lukas live, I understand your point Jonas. Fixed.

jonas-schievink: Hmm, is the diff wrong or are they still listed as inference roots?

src/types-and-traits.rst
```diff
@@ -2101,31 +2101,34 @@ Type Inference
...
+
 * :dp:`fls_A1NVYkepoaMk`
   The :t:`size operand` of an :t:`array expression` or an :t:`array type`.
 
+* :dp:`fls_cPlCLGCcl7EK`
+  A :t:`static initializer`.
```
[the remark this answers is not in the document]
kirtchev-adacore: I tend to alphabetize lists :rofl: just ask Lukas. I will co-locate constant initializer and static initializer.

src/types-and-traits.rst
```diff
@@ -2101,12 +2101,12 @@ Type Inference
 :t:`[expression]s` and :t:`[pattern]s` within a :t:`type inference root`.
 
 :dp:`fls_ybvrhh96fc7y`
-A :t:`type inference root` is an :t:`expression` whose inner :t:`[expression]s`
+A :t:`type inference root` is a :t:`construct` whose inner :t:`[expression]s`
```
jonas-schievink, line 2104: Is there a benefit to generalizing this to constructs instead of just expressions? Inference deals mostly with expressions and patterns, and all of the inference roots are expressions.

kirtchev-adacore: Yes, constant initializers, discriminant initializers, and static initializers have a `=` in their productions, so technically those are not expressions. This was caught by Lukas. :smile:

jonas-schievink (approved):

kirtchev-adacore:
bors merge

PR #384, opened 2022-11-10
Outline the type inference algorithm

jonas-schievink:
- Change *construct* back to *expression* by clarifying some rules about inference roots. This makes the description of the inference algorithm easier, since it only has to deal with expressions, statements, and patterns, and no other constructs.
- Outline the general inference algorithm, including all type fallback mechanisms present in current Rust.
- Write type inference rules for statements.

Next steps (to be done in later PRs):
- Write type inference rules for expressions (including coercions, higher-ranked inference, built-in arithmetic inference shortcuts, etc :dizzy_face:).
- Write typing rules for patterns.

### DIFF
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -2101,24 +2101,24 @@ Type Inference
 :t:`[expression]s` and :t:`[pattern]s` within a :t:`type inference root`.
 
 :dp:`fls_ybvrhh96fc7y`
-A :t:`type inference root` is a :t:`construct` whose inner :t:`[expression]s`
+A :t:`type inference root` is an :t:`expression` whose inner :t:`[expression]s`
 and :t:`[pattern]s` are subject to :t:`type inference` independently of those
 found in other :t:`[type inference root]s`.
 
 :dp:`fls_EWBilpepaDcX`
-The following :t:`[construct]s` are considered :t:`[type inference root]s`:
+The following :t:`[expression]s` are considered :t:`[type inference root]s`:
 
 * :dp:`fls_NYSzcvf5nQpi`
   A :t:`constant argument`.
 
 * :dp:`fls_htLp5J5ObgNh`
-  A :t:`constant initializer`.
+  The :t:`expression` of a :t:`constant initializer`.
 
 * :dp:`fls_cPlCLGCcl7EK`
-  A :t:`static initializer`.
+  The :t:`expression` of a :t:`static initializer`.
 
 * :dp:`fls_KphY5qHev0Dc`
-  The :t:`discriminant initializer` of an :t:`enum variant`.
+  The :t:`expression` of a :t:`discriminant initializer`.
 
 * :dp:`fls_Sowatt1V988J`
   A :t:`function body`.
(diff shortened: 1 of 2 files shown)


src/types-and-traits.rst
```diff
@@ -2182,16 +2184,70 @@ to :t:`[floating-point type]s`.
...
+:t:`[type]s` of :t:`[expression]s` and :t:`[pattern]s` across the
+:t:`type inference root` being inferred. In the rules detailed below, a static
+error occurs when :t:`type unification` fails.
+
+:dp:`fls_SZgixDCAx6PQ`
+:t:`Type inference` for a :t:`type inference root` proceeds as follows:
+
+#. :dp:`fls_XYY1U9h9HlAa`
+   Recursively process all :t:`[expression]s` and :t:`[statement]s` in the
+   :t:`type inference root` in program order.
```
kirtchev-adacore, line 2198:
```suggestion
   :t:`type inference root` in declaration order.
```
We have used that term before.

src/types-and-traits.rst
```diff
@@ -2182,16 +2184,70 @@ to :t:`[floating-point type]s`.
...
+   unifying them with ``i32``.
+
+#. :dp:`fls_ZFQhOxO3jpby`
+   If there are any remaining :t:`[floating-point type variable]s` that have not
+   been unified with a concrete :t:`floating-point type`, perform floating-point
+   type fallback by unifying them with ``f64``.
+
+#. :dp:`fls_l1G52a0qqEes`
+   If there are any remaining :t:`[diverging type variable]s` that have not been
+   unified with a concrete :t:`type`, unify them with ``()``.
```
kirtchev-adacore, line 2218:
```suggestion
   unified with a concrete :t:`type`, unify them with the :t:`unit type`.
```

src/types-and-traits.rst
```diff
@@ -2135,21 +2135,23 @@ depending on the :t:`type inference root` as follows:
...
-  The :t:`expected type` of the :t:`discriminant initializer` of an
-  :t:`enum variant` is determined as follows:
+  The :t:`expected type` of the :t:`expression` of a
+  :t:`discriminant initializer` is determined as follows:
 
   * :dp:`fls_vYvumjTQH9Xg`
-    If the :t:`enum type` is subject to :t:`attribute` :c:`repr` that specifies
-    a :t:`primitive representation`, the :t:`expected type` is the specified
-    :t:`primitive type`.
+    If the :t:`enum type` containing the :t:`discriminant` is subject to
```
kirtchev-adacore, line 2152:
```suggestion
    If the :t:`enum type` that contains the :t:`discriminant` is subject to
```

src/types-and-traits.rst
```diff
@@ -2182,16 +2184,70 @@ to :t:`[floating-point type]s`.
...
+
+:dp:`fls_hISRWZUuqE4Q`
+The :t:`type inference` rules for :t:`[statement]s` are as follows:
+
+* :dp:`fls_ygi1ACJ0RkfS`
+  :t:`[Item statement]s` are not subject to :t:`type inference`.
+
+* :dp:`fls_97Fxlv2KN6QF`
+  :t:`[Expression statement]s` apply the :t:`expression` inference rules below
+  to the contained :t:`expression`, without any :t:`expected type` set.
```
kirtchev-adacore, line 2232:
```suggestion
  to the related :t:`expression`, without any :t:`expected type` set.
```

kirtchev-adacore (changes requested):
I like the new rules!

src/types-and-traits.rst
kirtchev-adacore: Good point!

kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

src/types-and-traits.rst
jonas-schievink: Actually "evaluation order" is the better term here, "program order" typically means something else I believe

kirtchev-adacore: "Evaluation order" is a tricky one because in most cases it is "left-to-right", but in certain cases it is "right-to-left". Also, we have been using "declarative order" and "left-to-right order" in describing the dynamic semantics of evaluation and execution, so "evaluation order" seems to be a combination of "top-to-bottom" and "left-to-right" with the occasional "right-to-left".

PR #386, opened 2022-11-15
Rewrite the expression inference rules

jonas-schievink:
Rust itself has no concept of "type resolving" vs. "type imposing" expressions, so this replaces those concepts with a description closer to what Rust implementations actually do.

It also introduces the concept of incurring "obligations" during type inference, which I am not sure about. It matches what happens in practice, and describes the behavior of type inference fairly well, but we are already specifying requirements like "type X shall implement trait Y" everywhere, so it might be redundant.

### DIFF
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -2190,6 +2190,25 @@ The :t:`type inference` algorithm uses :t:`type unification` to propagate known
 :t:`type inference root` being inferred. In the rules detailed below, a static
 error occurs when :t:`type unification` fails.
 
+:dp:`fls_6GrNr2izovRN`
+Performing :t:`type inference` may introduce a requirement that some :t:`type`
+must implement a :t:`trait`, or that a :t:`type` or :t:`lifetime` must outlive
+some other :t:`lifetime`. Such requirements are referred to as
+:dt:`[obligation]s` and are detailed in the inference rules below.
+
+:dp:`fls_9dSltJ6U98Fo`
+If insufficient :t:`type` information is available at the time an
+:t:`obligation` is introduced, it may be deferred to be resolved later.
+Any time new :t:`type` information is derived during :t:`type inference`, the
+tool shall attempt to resolve all outstanding :t:`[obligation]s` and propagate
+any resulting :t:`type` information via :t:`type unification`.
+
+:dp:`fls_v5dWGuBKvQSJ`
+When an :t:`associated type` ``<Type as Trait>::Assoc`` is referenced within a
+:t:`type inference root` (either explicitly within the source code, or via the
+inferece rules below), an :t:`obligation` requiring that ``Type`` implements
+``Trait`` is introduced.
+
 :dp:`fls_SZgixDCAx6PQ`
 :t:`Type inference` for a :t:`type inference root` proceeds as follows:
 
@@ -2221,6 +2240,10 @@ error occurs when :t:`type unification` fails.
    If there are any remaining :t:`[global type variable]s` that have not been
    unified with a concrete :t:`type`, raise a static error.
 
+#. :dp:`fls_Nbdth8N0PSRq`
+   If there are any remaining :t:`[obligation]s` that do not hold or cannot be
+   resolved with the available :t:`type` information, raise a static error.
+
 :dp:`fls_hISRWZUuqE4Q`
 The :t:`type inference` rules for :t:`[statement]s` are as follows:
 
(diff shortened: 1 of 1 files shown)


src/types-and-traits.rst
```diff
@@ -2190,6 +2190,25 @@ The :t:`type inference` algorithm uses :t:`type unification` to propagate known
...
+Performing :t:`type inference` may introduce a requirement that some :t:`type`
+must implement a :t:`trait`, or that a :t:`type` or :t:`lifetime` must outlive
+some other :t:`lifetime`. Such requirements are referred to as
+:dt:`[obligation]s` and are detailed in the inference rules below.
+
+:dp:`fls_9dSltJ6U98Fo`
+If insufficient :t:`type` information is available at the time an
+:t:`obligation` is introduced, it may be deferred to be resolved later.
+Any time new :t:`type` information is derived during :t:`type inference`, the
+tool shall attempt to resolve all outstanding :t:`[obligation]s` and propagate
```
kirtchev-adacore, line 2203:
```suggestion
tool attempts to resolve all outstanding :t:`[obligation]s` and propagate
```
"shall" is for requirements aimed at the programmer.

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+
+* :dp:`fls_0HHC1iOk5dwz`
+  An :t:`if expression` is inferred by inferring its :t:`subject expression`
+  with an :t:`expected type` of :c:`bool`, then inferring its
+  :t:`block expression` with the :t:`expected type` of the :t:`if expression`.
+  Then, if the :t:`if expression` has an :t:`else expression`, apply the
+  inference rules below to it.
+
+* :dp:`fls_QZWTS0Giy3I3`
+  An :t:`if-let expression` is inferred by inferring its
```
kirtchev-adacore, line 2295:
```suggestion
  An :t:`if let expression` is inferred by inferring its
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+  :t:`subject let expression` with the :t:`expected type` set to the :t:`type`
+  of its :t:`pattern`, then inferring its :t:`block expression` with the
+  :t:`expected type` of the :t:`if-let expression`. If the
+  :t:`if-let expression` has an :t:`else expression`, apply the inference rules
```
kirtchev-adacore, line 2299:
```suggestion
  :t:`if let expression` has an :t:`else expression`, apply the inference rules
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+  below to it.
+
+* :dp:`fls_KJsIu1lgVZxP`
+  An :t:`else expression` that is part of an :t:`if expression` or
+  :t:`if-let expression` is inferred as follows:
```
kirtchev-adacore, line 2304:
```suggestion
  :t:`if let expression` is inferred as follows:
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+
+  * :dp:`fls_KRQxdSav1KBA`
+    If the :t:`else expression` has a :t:`block expression`, infer the
+    :t:`block expression` with the :t:`expected type` of the :t:`if expression`
+    or :t:`if-let expression`.
```
kirtchev-adacore, line 2309:
```suggestion
    or :t:`if let expression`.
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+
+  * :dp:`fls_Mcpwyvz47SoG`
+    If the :t:`else expression` has an :t:`if expression`, infer that nested
+    :t:`if expression` with the :t:`expected type` of the original
+    :t:`if expression`, then :t:`unify` its :t:`type` with the :t:`type` of
+    the original :t:`if expression` or :t:`if-let expression`.
```
kirtchev-adacore, line 2315:
```suggestion
    the original :t:`if expression` or :t:`if let expression`.
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+
+  * :dp:`fls_34AQ9g7xhdUj`
+    Otherwise, the :t:`else expression` has an :t:`if-let expression`. Infer
+    that nested :t:`if-let expression` with the :t:`expected type` of the
+    original :t:`if expression`, then :t:`unify` its :t:`type` with the
+    :t:`type` of the original :t:`if expression` or :t:`if-let expression`.
```
kirtchev-adacore, line 2321:
```suggestion
    Otherwise, the :t:`else expression` has an :t:`if let expression`. Infer
    that nested :t:`if let expression` with the :t:`expected type` of the
    original :t:`if expression`, then :t:`unify` its :t:`type` with the
    :t:`type` of the original :t:`if expression` or :t:`if let expression`.
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+     :t:`expected type` :c:`bool`.
+
+  #. :dp:`fls_F999gqcBfff9`
+     Infer the :t:`match arm body` of every :t:`match arm` with the
+     :t:`expected type` of the :t:`match expression`.
+
+* :dp:`fls_Esa4ST7lLp8T`
+  A :t:`for loop expression` is inferred by unifying the :t:`type` of its
+  :t:`pattern` with the :t:`type` ``<$T as core::iter::IntoIterator>::Item``,
+  where ``$T`` is the :t:`type` of the :t:`subject expression`, and then
```
kirtchev-adacore, line 2342:
```suggestion
  :t:`pattern` with the :t:`type` ``<T as core::iter::IntoIterator>::Item``,
  where ``T`` is the :t:`type` of the :t:`subject expression`, and then
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+         - ``!``
+         - :std:`core::ops::Not`
+       * - :dp:`fls_IoceMi7HfqsK`
+         - ``-``
+         - :std:`core::ops::Neg`
+
+  #. :dp:`fls_lDkPMB5UI58B`
+     Infer the :t:`type` of the :t:`expression` to be the :t:`associated type`
+     ``<$T as $Trait>::Output``, where ``$T`` is the :t:`type` of the
+     :t:`operand`, and ``$Trait`` is the operator :t:`trait` determined from
```
kirtchev-adacore, line 2378:
```suggestion
     ``<T as Trait>::Output``, where ``T`` is the :t:`type` of the
     :t:`operand`, and ``Trait`` is the operator :t:`trait` determined from
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+     If the :t:`expression` is neither a :t:`shift left expression` nor a
+     :t:`shift right expression`, and the :t:`expected type` is a
+     :t:`numeric type`, :t:`unify` the :t:`[type]s` of both :t:`[operand]s` with
+     the :t:`expected type`.
+
+  #. :dp:`fls_mCISAdm7sjRs`
+     Infer the :t:`type` of the :t:`expression` to be the :t:`associated type`
+     ``<$L as $Trait<$R>>::Output``, where ``$L`` is the :t:`type` of the
+     :t:`left operand`, ``$Trait`` is the operator :t:`trait` determined from
+     the table above, and ``$R`` is the :t:`type` of the :t:`right operand`.
```
kirtchev-adacore, line 2440:
```suggestion
     ``<L as Trait<R>>::Output``, where ``L`` is the :t:`type` of the
     :t:`left operand`, ``Trait`` is the operator :t:`trait` determined from
     the table above, and ``R`` is the :t:`type` of the :t:`right operand`.
```

src/types-and-traits.rst
```diff
@@ -2249,219 +2272,242 @@ The :t:`type inference` rules for :t:`[statement]s` are as follows:
...
+     :t:`trait` determined from the table above, and ``$R`` is the :t:`type` of
+     the :t:`modifying operand`.
+
+  #. :dp:`fls_0RZ7w0YqmzE3`
+     The :t:`type` of the :t:`expression` is the :t:`unit type`.
+
+* :dp:`fls_YppNCEPMYqWJ`
+  A :t:`comparison expression` is inferred by introducing an :t:`obligation`
+  ``$L: PartialEq<$R>``, where ``$L`` is the :t:`type` of the :t:`left operand`,
+  and ``$R`` is the :t:`type` of the :t:`right operand`. The :t:`type` of the
```
kirtchev-adacore, line 2497:
```suggestion
  A :t:`comparison expression` is inferred by introducing an :t:`obligation`
  ``L: PartialEq<R>``, where ``L`` is the :t:`type` of the :t:`left operand`,
  and ``R`` is the :t:`type` of the :t:`right operand`. The :t:`type` of the
```

kirtchev-adacore (changes requested):
I like these new rules!

By the way, it is a good idea to build the FLS and check that the links are sane. `./make.py -c` should do the trick.

src/types-and-traits.rst
jonas-schievink: hmm, most of our "shall" requirements are actually checked by the tool it seems

jonas-schievink:
> By the way, it is a good idea to build the FLS and check that the links are sane. ./make.py -c should do the trick.

I am doing that and the links look fine on my system. Is this about "if-let expression" vs. "if let expression"? It seems that the tooling canonicalizes both to the same reference, so either works.

kirtchev-adacore (approved):

PR #389, opened 2022-11-21
Evaluation of a closure expression does not evaluate the body

jonas-schievink:
Only calling the resulting value will evaluate the body


kirtchev-adacore (approved):
Indeed!

kirtchev-adacore:
bors merge

PR #392, opened 2022-11-22
Allow subtyping between assoc. consts in trait and impl

jonas-schievink:
For example, this is allowed:

```rust
trait Tr<'a> {
    const C: &'a str;
}

impl<'a> Tr<'a> for &'a () {
    const C: &'static str = "";
}
```


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #393, opened 2022-11-22
Document fundamental types/traits

jonas-schievink:
The `#[fundamental]` attribute is still unstable, so we don't specify it, but its behavior is observable and stable, so this PR documents that part.

Closes https://github.com/ferrocene/specification/issues/109


kirtchev-adacore (approved):

kirtchev-adacore:
bors merge

PR #395, opened 2022-11-23
Define unsafety without mentioning memory-safety guarantees

jonas-schievink:
Fixes https://github.com/ferrocene/specification/issues/158

All UB in Rust is ruled out statically, with the exception of UB caused by unsafe operations. Let's define them according to that.


Veykril (approved):

Veykril:
bors merge

PR #402, opened 2023-03-20
Remove the caution notice from the FLS

kirtchev-adacore:
The caution notice is no longer needed. The FLS already reflects 1.68.

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -36,9 +36,6 @@ templates_path = []
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = []
 
-rst_prolog = """
-"""
-
 
 # -- Options for HTML output -------------------------------------------------
 
--- a/src/conf.py
+++ b/src/conf.py
@@ -37,11 +37,6 @@ templates_path = []
 exclude_patterns = []
 
 rst_prolog = """
-.. caution::
-
-   You're reading a draft of the Ferrocene Language Specification. Some parts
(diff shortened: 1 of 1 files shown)


nava-j:
the fail seems to be a hash issue. It seems that livereload has been modified

kirtchev-adacore:
bors try

kirtchev-adacore:
bors try

src/conf.py
```diff
@@ -37,11 +37,6 @@
 exclude_patterns = []
 
 rst_prolog = """
-.. caution::
-
-   You're reading a draft of the Ferrocene Language Specification. Some parts
-   of this document might be missing, incomplete or incorrect. Our aim is to
-   have the specification ready by the end of 2022.
 """
```
pietroalbini, line 40: You should remove the whole prolog variable.

kirtchev-adacore: That was my initial reaction. But then I though, what if something else is using `rst_prolog`. I grepped within the repo and did not find anything that uses it, yet the variable is still needed. I found this a bit suspicious, so I decided to leave the variable in.

Let us see what the CI will say...

pietroalbini: `rst_prolog` is not present in the rest of the repository because it's [a Sphinx configuration variable](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_prolog) used by Sphinx itself. Since it contains no content it's safe to remove it.

pietroalbini (approved):
bors merge

PR #424, opened 2023-08-04
Various fixes

Veykril:
Closes https://github.com/ferrocene/specification/issues/111
Closes https://github.com/ferrocene/specification/issues/142
Closes https://github.com/ferrocene/specification/issues/166
Closes https://github.com/ferrocene/specification/issues/412

### DIFF
--- a/src/associated-items.rst
+++ b/src/associated-items.rst
@@ -91,6 +91,14 @@ within an :t:`implementation`.
 :dp:`fls_bx7931x4155h`
 An :t:`associated implementation type` shall have an :t:`initialization type`.
 
+:dp:`fls_bnTcCbDvdp94`
+An :t:`associated trait item` is an :t:`associated item` that appears
+within a :t:`trait`.
+
+:dp:`fls_N3cdn4lCZ2Bf`
+An :t:`associated trait implementation item` is an :t:`associated item` that
+appears within a :t:`trait implementation`.
+
 :dp:`fls_x564isbhobym`
 An :t:`associated trait constant` is an :t:`associated constant` that appears
 within a :t:`trait`.
@@ -130,6 +138,11 @@ is equivalent to a :t:`where clause` of the following form:
 
 :dp:`fls_oy92gzxgc273`
 A :t:`method` is an :t:`associated function` with a :t:`receiver`.
+
+:dp:`fls_oHxzyaiT7Qm6`
+The :t:`visibility modifier` of an :t:`associated trait item` or :t:`associated
+trait implementation item` is rejected, but may still be consumed by
+:t:`[macro]s`.
 
 .. rubric:: Examples
 
--- a/src/concurrency.rst
+++ b/src/concurrency.rst
@@ -32,9 +32,9 @@ Send and Sync
 .. rubric:: Legality Rules
 
 :dp:`fls_n5l17mlglq11`
-The Rust programming language provides synchronization facilities for
-:t:`[type]s` through the :std:`core::marker::Send` and
-:std:`core::marker::Sync` :t:`[trait]s`.
+The Rust programming language provides the :std:`core::marker::Send` and
+:std:`core::marker::Sync` :t:`[trait]s` for preventing data races at the
+:t:`type` level.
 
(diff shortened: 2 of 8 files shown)


src/associated-items.rst
```diff
@@ -91,6 +91,14 @@ within an :t:`implementation`.
 :dp:`fls_bx7931x4155h`
 An :t:`associated implementation type` shall have an :t:`initialization type`.
 
+:dp:`fls_bnTcCbDvdp94`
+An :t:`associated trait item` is an :t:`associated item` that appears
+within a :t:`trait`.
+
+:dp:`fls_N3cdn4lCZ2Bf`
+An :t:`associated trait implementation item` is an :t:`associated item` that
```
Dajamante, line 99: Would it be clearer to write a "trait implementation associated item"? Since the associated item is associated to the trait implementation? Or at least a dash: trait-implementation associated item?
Or is it against the standard (and we would need to change it everywhere!)

src/glossary.rst
```diff
@@ -464,6 +464,24 @@ associated trait function
 An :dt:`associated trait function` is an :t:`associated function` that appears
 within a :t:`trait`.
 
+.. _fls_47xtji9Pk8Lw:
+
+associated trait implementation item
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Dajamante, line 470: Yeah same here, I really think it is clearer to have trait-implementation associated item, at a glance.

src/lexical-elements.rst
```diff
@@ -343,7 +343,7 @@ The following names are used when referring to :t:`[punctuator]s`:
      - Equals
```
Dajamante, line 343: How are those operators selected, should we have all punctuation?
= assignment
or "//" comment?

src/values.rst
```diff
@@ -142,7 +142,8 @@ The :t:`type` of a :t:`static` shall implement the :std:`core::marker::Sized`
 :t:`trait`.
 
 :dp:`fls_doi4z6u55bi7`
-A :t:`mutable static` is a :t:`static` whose :t:`value` can be modified.
+A :t:`mutable static` is a :t:`static` with :t:`keyword` ``mut`` whose
```
Dajamante, line 145: Should we write it is global?

Dajamante (approved):

src/associated-items.rst
Veykril: we already have things like `associated trait constant` and such, so that's where this order comes from

src/concurrency.rst
```diff
@@ -32,9 +32,9 @@ Send and Sync
 .. rubric:: Legality Rules
 
 :dp:`fls_n5l17mlglq11`
-The Rust programming language provides synchronization facilities for
-:t:`[type]s` through the :std:`core::marker::Send` and
-:std:`core::marker::Sync` :t:`[trait]s`.
+The Rust programming language provides the :std:`core::marker::Send` and
+:std:`core::marker::Sync` :t:`[trait]s` for preventing data races at the
+:t:`type` level.
```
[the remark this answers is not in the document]
Veykril: what is more restrictive abou tthis? The main point about this phrasing is to get rid of the possible runtime intepretation from the previous wording. These traits have no runtime implications so care should be taken to word it accordingly

src/values.rst
Veykril: what does global as a term give us

src/lexical-elements.rst
Veykril: no I don't think we gain anything from adding this for all. The changes here were for completeness sake as the other operators of a similar group have the token names and semantic aliases

src/associated-items.rst
tshepang: I understand it as "trait constant" being the main thing, then pre-fixing it with "associated" to be more specific. "trait associated constant" would not flow as well.

Dajamante: The 3 word terms are ok, it is mostly the 4 word terms that become unclear.

Dajamante:
bors merge

PR #443, opened 2023-08-14
Fix up remaining unresolved links

Veykril:
There is one more unresolved thing (besides the broken paragraph linkes) which will be resolved by https://github.com/ferrocene/specification/issues/442


Dajamante (approved):

Dajamante:
bors merge

PR #446, opened 2023-09-07
Correct undefined behavior specification of ffi unwinding

Veykril:
Closes https://github.com/ferrocene/specification/issues/445

### DIFF
--- a/src/exceptions-and-errors.rst
+++ b/src/exceptions-and-errors.rst
@@ -58,11 +58,9 @@ Invoking :t:`macro` :std:`core::panic` has the following runtime effects:
    Control flow of the current thread resumes execution by invoking the
    :t:`function` subject to :t:`attribute` :c:`panic_handler`.
 
-.. rubric:: Undefined Behavior
-
-:dp:`fls_krhb07mvojph`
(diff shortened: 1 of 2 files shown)


src/exceptions-and-errors.rst
```diff
@@ -58,11 +58,9 @@ Invoking :t:`macro` :std:`core::panic` has the following runtime effects:
...
    :t:`function` subject to :t:`attribute` :c:`panic_handler`.
 
-.. rubric:: Undefined Behavior
-
-:dp:`fls_krhb07mvojph`
-It is undefined behavior when a :t:`panic` crosses a
-:t:`foreign function interface` boundary.
+#. :dp:`fls_xmtt04lw517w`
+   The process :t:`[abort]s` after control flow has been yielded by the
+   :t:`function` subject to :t:`attribute` :c:`panic_handler`.
```
pietroalbini, line 63: I don't think we should say that the process aborts. The panic handler type `fn(&PanicInfo) -> !`, so the panic handler never yields control flow.

pietroalbini (approved):
bors merge

PR #469, opened 2023-11-16
Fix struct pattern irrefutability to require the deconstructee to be a struct or union or single variant enum

Veykril:
Fixes https://github.com/ferrocene/specification/issues/455


pietroalbini (approved):
bors merge

PR #474, opened 2023-11-16
Remove paragraph about sequential action re-ordering

Veykril:
Fixes https://github.com/ferrocene/specification/issues/461

The issue seems correct to me here, this paragraph does not really make sense for Rust.


pietroalbini (approved):
bors merge

PR #496, opened 2024-06-07
remove empty field list in unions

nefelitav:
Rust unions cannot have zero fields, in contrast to enums and structs. (https://github.com/ferrocene/ferrocene/blob/main/tests/ui/union/union-empty.rs)

### DIFF
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -800,8 +800,10 @@ Union Types
 .. syntax::
 
    UnionDeclaration ::=
-       $$union$$ Name GenericParameterList? WhereClause? RecordStructFieldList
+       $$union$$ Name GenericParameterList? WhereClause? UnionFieldList
 
+   UnionFieldList ::=
+       $${$$ RecordStructField ($$,$$ RecordStructField)* $$,$$? $$}$$
 .. rubric:: Legality Rules
 
 :dp:`fls_nskmnzq95yqm`
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -800,15 +800,17 @@ Union Types
 .. syntax::
 
    UnionDeclaration ::=
-       $$union$$ Name GenericParameterList? WhereClause? UnionFieldList
+       $$union$$ Name GenericParameterList? WhereClause? RecordStructFieldList
 
-   UnionFieldList ::=
-       $${$$ RecordStructField ($$,$$ RecordStructField)* $$,$$? $$}$$
 .. rubric:: Legality Rules
 
 :dp:`fls_nskmnzq95yqm`
 A :t:`union type` is an :t:`abstract data type` that is a sum of other
 :t:`types`.
+
+:dp:`fls_nskmnzq95yqn`
+A :t:`union` without any :t:`[union field]s` is rejected, but may still be consumed by
+:t:`[macro]s`.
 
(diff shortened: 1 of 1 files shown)


Veykril:
Good catch! Though this is a semantic restriction, not a syntactic one (that is macros can observe the syntax still), as can be seen by the following compiling https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

Could you phrase this as a legality rule? (like is done for `unsafe` tokens for modules in https://spec.ferrocene.dev/program-structure-and-compilation.html#fls_whgv72emrm47)

nefelitav:
> Good catch! Though this is a semantic restriction, not a syntactic one (that is macros can observe the syntax still), as can be seen by the following compiling https://play.rust-lang.org/?version=stable&mode=debug&edition=2021
> 
> Could you phrase this as a legality rule? (like is done for `unsafe` tokens for modules in https://spec.ferrocene.dev/program-structure-and-compilation.html#fls_whgv72emrm47)

Thank you for the feedback and the clarification! It seems there might have been an issue with the link you provided on the playground. Could you please resend it? 

Also, based on my understanding of the Rust Reference, it seems like the [Union ](https://doc.rust-lang.org/stable/reference/items/unions.html)Syntax requires at least one field, in contrast to [Struct ](https://doc.rust-lang.org/stable/reference/items/structs.html)and [Enum](https://doc.rust-lang.org/stable/reference/items/enumerations.html). Could you please confirm if this is correct?

Veykril:
Whoops, here is the link https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a3c7c2e0cafb924ead10ded4dd93a4b1

Veykril:
Reading through the reference, it seems to actually not describe this restriction at all. The grammar lists `StructFields` like we do, but  they are also lacking a phrasing that requires at least on field in the declaration.

nefelitav:
> Reading through the reference, it seems to actually not describe this restriction at all. The grammar lists `StructFields` like we do, but they are also lacking a phrasing that requires at least on field in the declaration.

Thank you for the link! I apologize in advance if I'm wrong, but I still think the Reference syntax is different from the Ferrocene one. In the reference, for [unions](https://doc.rust-lang.org/stable/reference/items/unions.html?highlight=StructFields), there is no "?" after the StructFields, while for [structs ](https://doc.rust-lang.org/stable/reference/items/structs.html?highlight=StructFields) and [enums ](https://doc.rust-lang.org/stable/reference/items/enumerations.html?highlight=structfields) there is a "?", showing that fields are optional.  And StructFields are defined like this:

StructFields :
   StructField (, StructField)* ,?
   
 So, I think that according to the Reference syntax, there should be at least one StructField in unions .

Veykril:
Ah I see, I didn't look at the definition of `StructFields` properly, in that case the reference is wrong given you can have zero fields in the grammar as shown by my snippet.

nefelitav:
> Ah I see, I didn't look at the definition of `StructFields` properly, in that case the reference is wrong given you can have zero fields in the grammar as shown by my snippet. It is a semantic rule, not syntactic, so both documents are missing the semantic rule disallowing this state post macro expansion.

Thank you for the explanation! I hope it is okay now.

Veykril (approved):

Veykril:
Sounds good, thanks!
bors merge

PR #497, opened 2024-06-07
Fix typos

nefelitav:
use "a" before consonant sound and "an" before vowel sound


Urhengulas:
Thank you @nefelitav :D

PR #498, opened 2024-06-13
Update to 1.77

Veykril:


src/lexical-elements.rst
```diff
@@ -703,6 +704,112 @@ The :t:`type` of a :t:`raw byte string literal` of size ``N`` is ``&'static
...
+
+.. rubric:: Legality Rules
+
+:dp:`fls_nPI7j0siGP8G`
+A :t:`simple c string literal` is a :t:`c string literal` where the characters are
+:t:`Unicode` characters.
+
+:dp:`fls_Ae7LM4Wg0NA7`
+The :t:`type` of a :t:`simple string literal` is :std:`&'static
+[core::ffi::CStr]`.
```
pietroalbini, line 760: Shouldn't this be `&'static core::ffi::CStr`?

src/lexical-elements.rst
```diff
@@ -703,6 +704,112 @@ The :t:`type` of a :t:`raw byte string literal` of size ``N`` is ``&'static
...
+
+.. rubric:: Legality Rules
+
+:dp:`fls_gLrei65i8Uzq`
+A :t:`raw c string literal` is a :t:`simple c string literal` that does not
+recognize :t:`[escaped character]s`.
+
+:dp:`fls_9nJHsg9dCi66`
+The :t:`type` of a :t:`simple string literal` is :std:`&'static
+[core::ffi::CStr]`.
```
pietroalbini, line 803: Shouldn't this be `&'static core::fii::CStr`?

pietroalbini (changes requested):

src/lexical-elements.rst
Veykril: The `[]` is the processing syntax for the std linkage, it does render as `&'static core::ffi::CStr`

Veykril:
Hmm, for some reason the `CStringLiteral` syntax rule is not being linked to

Veykril:
Seems to be a bug in our tooling, will investigate that later, should not block this PR.

src/lexical-elements.rst
pietroalbini: ...right :sweat_smile:

pietroalbini (approved):
bors merge

PR #502, opened 2024-06-16
add while loop syntax fix

nefelitav:
According to the Rust Reference, "while loop expressions", as well as "if expressions" should accept any expression except struct expressions (SubjectExpression). So, this must be more accurate.


Veykril (approved):

Veykril:
Thanks!
bors merge

PR #503, opened 2024-06-17
Monkey patch sphinx SortIDs transform to sort our own IDs

Veykril:
This fixes some anchoring inconsistencies and also makes link check correctly work for section id <-> other id collisions, see 3cf79ed8b0c7afc4d602cec5e9e7fb776102dd30 which CI did not catch so far, now it will. cc https://github.com/ferrocene/ferrocene/pull/699 which broke because of the collision

### DIFF
--- a/exts/ferrocene_spec/__init__.py
+++ b/exts/ferrocene_spec/__init__.py
@@ -3,6 +3,7 @@
 
 from . import definitions, informational, syntax_directive, std_role, paragraph_ids
 from . import items_with_rubric
+from .utils import FlsSortIds
 from sphinx.domains import Domain
 
 
@@ -37,6 +38,7 @@ class SpecDomain(Domain):
 
 def setup(app):
     app.add_domain(SpecDomain)
+    app.add_transform(FlsSortIds)
     definitions.setup(app)
     paragraph_ids.setup(app)
     informational.setup(app)
--- a/exts/ferrocene_spec/utils.py
+++ b/exts/ferrocene_spec/utils.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: The Ferrocene Developers
 
 from docutils import nodes
+from sphinx import transforms
 
 
 def section_id_and_anchor(section):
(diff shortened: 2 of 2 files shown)


exts/ferrocene_spec/__init__.py
```diff
@@ -62,3 +63,27 @@ def setup(app):
...
+# into `fls-{id}` which can break the link check from working correctly.
+def path_id_sorting():
+    def sort_and_patch_underscore(self, **kwargs) -> None:
+        from docutils import nodes
+
+        for node in self.document.findall(nodes.section):
+            for n in range(len(node["ids"])):
+                id = node["ids"][n]
+                if id.startswith("fls-"):
+                    node["ids"][n] = id[:3] + "_" + id[4:]
```
Urhengulas, line 81:
```suggestion
                    node["ids"][n] = "fls_" + id.removeprefix("fls-")
```

exts/ferrocene_spec/__init__.py
```diff
@@ -62,3 +63,27 @@ def setup(app):
...
+# So we patch the transform instead so that our fls ids are sorted to the back.
+# In addition to that we normalize them, as sphinx turns the `_` in `fls_{id}`
+# into `fls-{id}` which can break the link check from working correctly.
+def path_id_sorting():
+    def sort_and_patch_underscore(self, **kwargs) -> None:
+        from docutils import nodes
+
+        for node in self.document.findall(nodes.section):
+            for n in range(len(node["ids"])):
+                id = node["ids"][n]
```
Urhengulas, line 79:
```suggestion
            for n, id in enumerate(node["ids"]):
```

exts/ferrocene_spec/__init__.py
Urhengulas: Or
```suggestion
                    node["ids"][n] = id.replace("-", "_", 1)
```

Veykril:
CI fails due to https://github.com/ferrocene/specification/pull/505

pietroalbini:
I'm not too thrilled with the monkey patching approach. If we want to guarantee that the FLS ID is always the first ID, we should add our own transform that moves it to the front.

pietroalbini (approved):
bors merge

PR #506, opened 2024-06-19
clarify when each field selector is used

nefelitav:
it should be clear that the indexed field selector is used only with tuples and the named field selector is only used with structs and unions.


Veykril:
This is already implied/described in the resolutions chapter (where this restriction makes more sense to be described imo) https://spec.ferrocene.dev/entities-and-resolution.html#field-resolution

PR #507, opened 2024-06-19
explain union behavior in writing

nefelitav:
I think it is important to explain this behavior of unions, that when writing to one field, the others are overwritten. I couldn't see it mentioned somewhere.


Veykril:
this is already described in https://spec.ferrocene.dev/types-and-traits.html#fls_gt3tkbn4bsa6

PR #513, opened 2024-07-05
Specify auto trait trait object upcasting

Veykril:


src/types-and-traits.rst
```diff
@@ -1156,10 +1156,14 @@ Trait Object Types
 A :t:`trait object type` is a :t:`type` that implements a :t:`trait`, where the
 :t:`type` is not known at compile time.
 
+:dp:`fls_eWac7zOda3lh`
+The :t:`principal trait` of :t:`trait object type` is the first :t:`trait bound`.
```
Veykril, line 1160: This definition isn't used not (thought I'd need the term when describing this at first), but it will become relevant in the future when full upcasting exists

pietroalbini (approved):
bors merge

PR #518, opened 2024-09-03
Define ShiftLeftExpression and ShiftRightExpression evaluation for integer types

### LINKED ISSUE #517: Missing bits in Shift{Left,Right}Expression definition
(pun intended)

It looks like the definition of `ShiftLeftExpression` and `ShiftRighExpression` is incomplete.

More precisely, it does not define what a left (respectively right) shift evaluates to when both operands in an [Integer Type]s. The only thing it says its value is `core::ops::Shl::shl(left_operand, right_operand)` (resp. `shr`).

I quickly at the definition of [`Shl`] (resp. [`Shr`]) and found no clear definition either.

I would expect something in the lines of:

> If the type of both operands is an integer type, then the left shift (resp. right) expression evaluates to [...].

IMO the following should be covered:

- Bits are... well... shifted
- What happens when the LHS has a negative value (`rustc` extends the sign bit for `>>` and does nothing special for `<<`)
- What happens when the RHS is negative (`rustc` panics in `debug` mode and... always returns `-1` or `0` in release mode?)

Thanks for building Ferrocene!

[Integer Type]: https://spec.ferrocene.dev/types-and-traits.html#integer-types
[`Shl`]: https://public-docs.ferrocene.dev/main/core/ops/trait.Shl.html
[`Shr`]: https://public-docs.ferrocene.dev/main/core/ops/trait.Shr.html

scrabsha:
Fix #517.

I did my best to follow the style of the rest of the specification.

Thank you for building Ferrocene!!

### DIFF
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -1719,7 +1719,17 @@ The :t:`evaluation` of a :t:`shift left expression` proceeds as follows:
    The :t:`right operand` is evaluated.
 
 #. :dp:`fls_f0p70y92k14f`
-   ``core::ops::Shl::shl(left_operand, right_operand)`` is invoked.
+   If the types of both :t:`[operand]s` are :t:`[integer type]s`, then the
+   :t:`shift left expression` evaluates to the value of the left :t:`operand`
+   whose bits are shifted left by the number of positions the right :t:`operand`
+   evaluates to. Vacated bits are filled with zeros. ``lhs << rhs`` evaluates
+   to :math:`\mathrm{lhs} \times 2 ^ \mathrm{rhs}`, casted to the type of the left
+   :t:`operand`. If the value of the right :t:`operand` is negative or greater
+   than or equal to the width of the left operand, then the operation results in
+   an :t:`arithmetic overflow`.
+
+#. :dp:`fls_8QGbl2SBU3R0`
+   Otherwise, ``core::ops::Shl::shl(left_operand, right_operand)`` is invoked.
 
 :dp:`fls_303r0u6ug215`
 The :t:`evaluation` of a :t:`shift right expression` proceeds as follows:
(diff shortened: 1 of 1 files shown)


src/expressions.rst
```diff
@@ -1719,7 +1719,23 @@ The :t:`evaluation` of a :t:`shift left expression` proceeds as follows:
    The :t:`right operand` is evaluated.
 
 #. :dp:`fls_f0p70y92k14f`
-   ``core::ops::Shl::shl(left_operand, right_operand)`` is invoked.
+   If the type of the left :t:`[operand]` is any :t:`[unsigned integer type]`
+   and the type of the right :t:`[operand]` is any :t:`[integer type]`, then
+   the :t:`[shift left expression]` evaluates to the left logical shift of the
```
Veykril, line 1724: This introduces the concept of a left logical shift which is not defined (likewise for arithmetic below)

src/expressions.rst
```diff
@@ -5099,9 +5131,10 @@ Arithmetic Overflow
 -------------------
 
 :dp:`fls_oFIRXBPXu6Zv`
-An :t:`arithmetic overflow` occurs when an :t:`arithmetic expression` or a
-:t:`negation expression` computes a :t:`value` of a :t:`scalar type` that lies
-outside of the range of valid :t:`[value]s` for the :t:`scalar type`.
+An :t:`arithmetic overflow` occurs when an :t:`arithmetic expression`, a
+:t:`negation expression` or a :t:`bit expression` computes a :t:`value` of a
+:t:`scalar type` that lies outside of the range of valid :t:`[value]s` for the
+:t:`scalar type`.
```
Veykril, line 5137: We can generalize this by just stating `operator expression`

Veykril (approved):

Veykril:
Thanks!
bors merge

PR #519, opened 2024-09-09
Ignore VSCode settings

Veykril:


.gitignore
```diff
@@ -11,3 +11,6 @@ __pycache__
 
 # Linkchecker ignored files
 /.linkchecker
+
+# VSCode ignored files
+/.vscode
```
pietroalbini, line 16: You can add this to `.git/info/exclude` to ignore it in your local checkout without committing the ignore to the repo.

Veykril: Even better, just realized I can ignore `.vscode/settings.json` globally

PR #521, opened 2024-09-09
Update changelog for 1.80.0

Veykril:


src/values.rst
```diff
@@ -37,6 +48,20 @@ Two :t:`[value]s` :t:`overlap` when
...
+
+:dp:`fls_uhwpuv6cx4ip`
+It is undefined behavior to create an :t:`allocated object` with :t:`memory
+size` ``size`` at a :t:`base address` ``base`` where ``base + size`` is greater
+than the architectures maximum :c:`usize` value.
+
+:dp:`fls_xuuFKmm181bs`
+It is undefined behavior to create an :t:`allocated object` with :t:`memory
+size` ``size`` where ``size`` is greater than the architectures maximum
+:c:`isize` value.
```
pietroalbini, line 63: This should be in an undefined behavior rubric, so that it shows up in the UB appendix.

pietroalbini (changes requested):

src/values.rst
Veykril: It already is (see line 45)

pietroalbini: Whoops!

pietroalbini (approved):
bors merge

PR #530, opened 2024-11-04
Fix array repetition operand evaluation

scrabsha:
It turns out that `[<repeated>; <length>]` evaluates `<repeated>` only one time, as seen in the following example:

```rust
fn main() {
    let mut counter = 0;
    [{
        counter += 1;
    }; 9];
    assert_eq!(counter, 1);
}
```

###### [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b7ca3de9ba64675238380acae97fde37)

This PR updates the doc in order to match this behavior.


Veykril (approved):

Veykril:
Thanks!
bors merge

PR #541, opened 2025-01-16
Mention the rust version the spec targets

Veykril:
Will require changes in ferrocene when this gets mirrored I think?

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -19,6 +19,8 @@ author = "The Ferrocene Developers"
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


pietroalbini:
In `ferrocene/ferrocene` the extension part is already fully implemented and hooked up. Can we just add the substitution in the monorepo?

Veykril:
The build here will fail though if we don't set the substitution locally (and we can't hook up with substitution tooling from `ferrocene_qualification` given it defines a bunch of extra required args)

pietroalbini (approved):
bors merge

PR #542, opened 2025-01-21
Convert the FLS to pull sphinx shared resources from pypi

pietroalbini:
This PR changes the FLS to use `uv` to manage its Python dependencies, and to pull our shared resources from PyPI instead of the submodule.


Veykril:
bors merge

PR #543, opened 2025-01-21
Disable dependabot for submodules

emilyalbini:
We are not using submodules anymore. [Support for uv is not yet implemented, but is planned this quarter](https://www.github.com/dependabot/dependabot-core/issues/10478).


Dajamante (approved):
bors merge

PR #547, opened 2025-02-04
Remove invalid dummy

Veykril:
It was meant as a future TODO, but what it wants to descibe is not reasonable

Closes https://github.com/ferrocene/specification/issues/545


pietroalbini (approved):
bors merge

PR #552, opened 2025-04-17
describe 1.86 changes

tshepang:


src/types-and-traits.rst
```diff
@@ -2048,6 +2048,19 @@ occur when:
...
+
+  .. code-block:: rust
+
+     trait Sub: Super {}
+     trait Super {}
+
+     fn upcast(x: &dyn Sub) -> &dyn Super {
+         x // implicit coercion
+     }
+
```
Veykril, line 2063:
```suggestion
* :dp:`fls_QB4c6FNKxaPl`
  The source :t:`type` is a :t:`trait object type` with some :t:`principal trait bound` ``T`` and the target :t:`type` is a :t:`trait object type` with some :t:`principal trait bound` ``U``, where ``U`` is a :t:`supertrait` of ``T``.

```

Note that we don't throw in code example blocks arbitrarily unless describing the structure is fairly complicated (and even then I feel like it is wrong to do so)

tshepang:
re-submitted in https://github.com/rust-lang/fls/pull/566

PR #554, opened 2025-04-17
fix broken link

tshepang:


src/types-and-traits.rst
```diff
@@ -2046,7 +2046,7 @@ occur when:
   :t:`type`.
 
 * :dp:`fls_SYnFJBhi0IWj`
-  The source :t:`type` is a :t:`trait object type` and the target :t:`type` is a :t:`trait object type` with the same or no :t:`[principal trait bound]`, and the target :t:`type` has the same or less non-:t:`principal trait` :t:`[trait bound]s`.
+  The source :t:`type` is a :t:`trait object type` and the target :t:`type` is a :t:`trait object type` with the same or no :t:`[principal trait]`, and the target :t:`type` has the same or less non-:t:`principal trait` :t:`[trait bound]s`.
```
Veykril, line 2049:
```suggestion
  The source :t:`type` is a :t:`trait object type` and the target :t:`type` is a :t:`trait object type` with the same or no :t:`principal trait`, and the target :t:`type` has the same or less non-:t:`principal trait` :t:`[trait bound]s`.
```

The `[]` are unnecessary actually, given this is the correct linkable term already. They are used in case we modify the actual text, like pluralizing for example, ``:t:`[principal trait]s` ``

Veykril (approved):

Veykril:
bors merge

Issue #155, opened 2022-07-26
Glossary: Outdated Unicode ISO Standard reference


ojeda:
ISO/IEC 10646:2017 is not the latest edition, 2020 is.

PLeVasseur:
Possibly addressed by the linked, merged PR. @tshepang -- what do you think?

Also ping to @ojeda to see if that addresses the concern.

ojeda:
No, this issue was about the Glossary, and from a quick look, it still refers to the older version. There are now amendments to the 2020 one from a quick look.

Having said that, given https://github.com/rust-lang/fls/issues/156, it may be simplest to avoid to refer to the ISO/IEC one to begin with, especially if Rust itself refers to the Unicode Standard. Plus the ISO/IEC may lag behind the other one anyway, as far as I understand.

ojeda:
I sent a quick PR with a suggestion for both issues: https://github.com/rust-lang/fls/pull/617

PR #557, opened 2025-05-06
Claim to follow Unicode 16 for lexing identifiers.

mattheww:
Rustc has used Unicode 16 since Rust 1.83.

This change brings the FLS into sync with the Reference for this detail.
See rust-lang/reference#1688.

I'm not sure whether the changelog should be updated (and if so, how). The FLS was previously claiming Unicode version 13. I don't have notes saying which versions the updates to Unicode 14 and 15 happened in.

Unicode normalization is also dependent on the Unicode version, but there's no version number in the existing text of §2.3:16 (`fls_vde7gev5rz4q`).

### DIFF
--- a/src/lexical-elements.rst
+++ b/src/lexical-elements.rst
@@ -476,7 +476,7 @@ keyword]s`.
 
 :dp:`fls_jpecw46eh061`
 A :t:`pure identifier` shall follow the specification in Unicode Standard Annex
-#31 for :t:`Unicode` version 13.0, with the following profile:
+#31 for :t:`Unicode` version 16.0, with the following profile:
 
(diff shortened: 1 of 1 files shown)


traviscross:
cc @Veykril

Veykril:
The changelog (as used so far) generally only tracks the language level changes from the RELEASES.md, so if that was not listed there it doesn't need to touch the changelog I believe.

mattheww:
See also #155

traviscross:
cc @tshepang @plevasseur

tshepang:
whether this should be mentioned on changelog can be decided without blocking the clear improvement

traviscross (approved):

PR #558, opened 2025-05-09
values: fix paragraph on memory address of an allocated object

tshepang:


src/values.rst
```diff
@@ -18,8 +18,7 @@ be stored in a memory location, and interpreted based on some :t:`type`.
 An :dt:`allocated object` is a :t:`value` stored at some memory address.
 
 :dp:`fls_kaomYy0Ml4Nh`
-An :t:`[allocated object]s` :dt:`base address` is the the memory address the
-object is stored.
+The :dt:`base address` of an :t:`allocated object` is the memory address where the object is stored.
```
traviscross, line 21: The FLS is hard wrapped at 80 characters.  It seems worth being consistent with that until we decide to do otherwise.

Veykril: > (for the most part)

It started at a 80 character hard wrap, but I was trying to push for breaking after sentence end instead some time after which is why its a bit mixed at the moment. Given the transfer now thats something to make a decision on probably (and then re-enforce that across the document).

traviscross: Thanks for the background.  Probably we'd just unwrap paragraphs entirely then, as we decided to do in the Reference.

PR #561, opened 2025-05-15
add `else` as a keyword

rbakbashev:


tshepang:
sorry for the delay on this... could you fix the conflict

PR #563, opened 2025-05-30
Move external shared Ferrocence extensions locally to FLS Project

JoelMarcey:
Ferrocene has a shared resource package for their documentation and other sites that was used for the FLS: https://pypi.org/project/ferrocene-sphinx-shared-resources/. In order for us to make changes to the theme, logo, etc., I moved some of these extensions locally to the FLS Project since we wouldn't be able to make any changes to these things in that external package since it would affect other Ferrocene things too.

Now there are no dependencies on the external Ferrocene packages and the site builds as normal with them locally.


JoelMarcey:
Looks like CI is failing on license compliance with the locally imported files. Let me check into that.

JoelMarcey:
Ok. Looks like CI is happy now with the licenses.

exts/ferrocene_toctrees.py
```diff
@@ -0,0 +1,135 @@
...
+        if not isinstance(obj, TocTreeCollector):
+            continue
+        # When running sphinx-autobuild, this function might be called multiple
+        # times. When the collector is already disabled `listener_ids` will be
+        # `None`, and thus we don't need to disable it again.
+        #
+        # Note that disabling an already disabled collector will fail.
+        if obj.listener_ids is None:
+            continue
+        obj.disable(app)
```
pietroalbini, line 122: I deeply apologize for this function.

make.py
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env -S uv run
 # SPDX-License-Identifier: MIT OR Apache-2.0
 # SPDX-FileCopyrightText: The Ferrocene Developers
+# SPDX-FileCopyrightText: The Rust Project Contributors
 
 import os
-import ferrocene_standalone_make_cli
+from src import ferrocene_standalone_make_cli
```
pietroalbini, line 7: Instead of copying the `ferrocene_standalone_make_cli` module into the repo and importing it, we should just replace `make.py` with the contents of that file, and adding a call to the `main` function at the end of it.

Originally this was a module inside of ferrocene-sphinx-shared-resources as we wanted to share the `make.py` implementation across multiple Sphinx sites, but we don't need that once we copy the file inside of this repo.

src/conf.py
```diff
@@ -27,6 +28,7 @@
     "ferrocene_spec",
     "ferrocene_spec_lints",
     "ferrocene_toctrees",
+    "ferrocene_theme",
```
pietroalbini, line 31: Themes don't have to be defined in packages, Ferrocene only did that to be able to distribute it on PyPI. If you move `exts/ferrocene_theme` to `themes/ferrocene` and add `html_theme_path = ["../themes"]` to `conf.py`, loading the theme will still work (but without an extension in the middle).

emilyalbini:
Looks good to me, the only change I ask is to reword the `Address @pietroalbini feedback` commit message to remove the `@`. If the `@` is there I will get an email notification every time someone pushes that commit to a repo, which might get annoying.

JoelMarcey:
> the only change I ask is to reword the Address @pietroalbini feedback commit message to remove the @

Done @pietroalbini . Thanks!

.gitignore
```diff
@@ -11,3 +11,6 @@ __pycache__
 
 # Linkchecker ignored files
 /.linkchecker
+
+# Mac OS ignored files
+.DS_Store
```
traviscross, line 16: This file is now missing a trailing newline.  In one of the follow-on PRs, @JoelMarcey, if you could fix this, that would be good.

PR #585, opened 2025-10-31
feat: add issue template for changes

PLeVasseur:
As we'd discussed, it seems worthwhile to have a way for folks to create issues quickly from a template for changes to the FLS.

This is a first step.

The next step I will use an issue auto labeler to apply labels based on choices made on dropdowns.

### DIFF
--- a/.github/ISSUE_TEMPLATE/change.yml
+++ b/.github/ISSUE_TEMPLATE/change.yml
@@ -66,9 +66,9 @@ body:
       options:
         - Bug
         - Formatting
-        - Improvement to Non-Prose Section
-        - Language Cleanup
-        - New Content
+        - Improvement to non-prose section
+        - Language cleanup
+        - New content
         - Other
     validations:
       required: true


tshepang:
I would assign copyright to "The Rust Project Developers", to match some of the other files here, and as a generic copyright holder text for Rust Project in general

PLeVasseur:
> I would assign copyright to "The Rust Project Developers", to match some of the other files here, and as a generic copyright holder text for Rust Project in general

Good call. Updated in https://github.com/rust-lang/fls/pull/585/commits/cefd25df5f079076390b3d37c545b364a011edbf

.github/ISSUE_TEMPLATE/change.yml
```diff
@@ -0,0 +1,99 @@
...
+    attributes:
+      label: Category
+      description: What kind of change is this?
+      options:
+        - Bug
+        - Formatting
+        - Improvement to Non-Prose Section
+        - Language Cleanup
+        - New Content
+        - Other
```
traviscross, line 72:
```suggestion
        - Bug
        - Formatting
        - Improvement to non-prose section
        - Language cleanup
        - New content
        - Other
```

Sentence-casing seems perhaps better for this.

.github/ISSUE_TEMPLATE/change.yml
```diff
@@ -0,0 +1,99 @@
...
+  - type: input
+    id: fls-id
+    attributes:
+      label: FLS Paragraph ID
+      description: |
+        The Paragraph ID from the FLS corresponding to what this Guideline covers.
+        Easiest way to obtain this is to navigate to the
+        [FLS](https://rust-lang.github.io/fls), right click a section ID
+        (e.g. `4.2:2`), inspect, and then find it in the pane which opens in
+        your browser.
```
traviscross, line 84: As an aside, perhaps this points toward us wanting to expose these identifiers more readily on the front-end.

.github/ISSUE_TEMPLATE/change.yml
```diff
@@ -0,0 +1,99 @@
...
+
+        **Note**: New content need not fill this in.
+      placeholder: fls_69zyas59o8ff
```
traviscross:
```suggestion
      placeholder: fls_deaddeadbeef
```

Maybe we want to use a more obvious placeholder here so we're more likely to notice if it's not changed.

PLeVasseur: This is a good idea.

.github/ISSUE_TEMPLATE/change.yml
PLeVasseur: Yeah, agreed. I think something like display on hover over the section numbering would be really nice.

PLeVasseur: Opened an issue for it: https://github.com/rust-lang/fls/issues/586

kirtchev-adacore:
The template looks reasonable to me, but then again I am no Github template expert :grin:

tshepang:
let's see how this gets used

Issue #623, opened 2025-11-26
[Change]: [1.91.0, 1.91.1] Lower pattern bindings in the order they're written and base drop order on primary bindings' order


PLeVasseur:
### Change description

Lower pattern bindings in the order they're written and base drop order on primary bindings' order

### Area

patterns

### Category

New content

### FLS Paragraph ID

N/A

### Change description (extended)

1.91 change log:
https://github.com/rust-lang/fls/pull/621

[Lower pattern bindings in the order they're written and base drop order on primary bindings' order](https://github.com/rust-lang/rust/pull/143764)

Reference PR:

- https://github.com/rust-lang/reference/pull/1953

PR #631, opened 2025-12-06
feat: add first draft of drop order in or patterns clarification

PLeVasseur:
### What this does

* Creates a new `or pattern` which is intended to match the [Reference's same concept](https://doc.rust-lang.org/reference/patterns.html#or-patterns)
* Uses new `or pattern` in order to clarify the ordering of evaluation of the `pattern-without-alternation`
* Adds an example of how the `or pattern` drop order works

### Still open, still thinking about it

I don't think I was able to get the definition of the `or pattern` to match 1-to-1 currently, as in the Reference
it refers to:
> let-bindings and function and closure arguments

I didn't seen these concepts defined in that same way. I think it's at least possible there are similar concepts
in the FLS I haven't found yet.

For example, I think potentially [`argument operand`](https://rust-lang.github.io/fls/glossary.html#argument-operand) could be used for referring to "function arguments".

Will need to give this a bit more thought, but open to hear from others.

closes https://github.com/rust-lang/fls/issues/623

Corresponding Reference PR:

- https://github.com/rust-lang/reference/pull/1953

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -4990,8 +4990,7 @@ pattern matching
 ^^^^^^^^^^^^^^^^
 
 :dp:`fls_y3oputy9e0sz`
-:dt:`Pattern matching` is the process of matching a :t:`pattern` against a
-:t:`value`.
+:t:`Pattern matching` is the process of matching a :t:`pattern` against a :t:`value`.
 
 .. _fls_cptagvgpgnze:
 
--- a/src/patterns.rst
+++ b/src/patterns.rst
@@ -1299,6 +1299,9 @@ follows:
 Pattern Matching
 ----------------
 
+:dp:`fls_zv73CR8rplIa`
+:dt:`Pattern matching` is the process of matching a :t:`pattern` against a :t:`value`.
+
 .. rubric:: Legality Rules
 
 :dp:`fls_tlwr4u7bjhh5`


kirtchev-adacore:
Hi Pete! Here is some feedback, which attempts to make the new text a bit more formal. Note that I am actually introducing more rules to cover the missing semantics for or-patterns.

**Changes to `src/patterns.rst`:**

I would put the definition of "or-pattern" immediately after that of "pattern", with the following new text:
```
An or-pattern is a pattern which or-s two or more subpatterns using character 0x7C (vertical line).
```
After `5:7 It is a static error` (since we are in the place where we impose some restrictions, from more general to more specific):
```
An or-pattern shall not appear in the pattern-without-alternation of a closure parameter, a function parameter, or a let statement.
```
(note that the Reference states "function and closure arguments", but I think they mean function and closure **parameters**)

Change `5:8, 5:9, 5:10` into:
```
Any two subpatterns of an or-pattern are subject to the following restructions:

* The types of the two subpatterns shall be unifiable.

* The set of bindings introduced by the two patterns shall be distinct,

* Any two bindings with the same name in the two patterns shall have unifiable types and shall have the same binding modes.
```
Insert the following text after `5:4:1 Pattern matching that involves`:
```
If the pattern is an or-pattern, then perform or-pattern matching.
```
Add new section `5.4.1. Or-pattern Matching` with the following text:
```
5.4.1. Or-pattern Matching

Or-pattern matching of an or-pattern of the form `constructor(or-pattern, rest)`, where `constructor` is an arbitrary constructor, `or-pattern` is the or-pattern, and `rest` is optionally a remaining pattern, proceeds as follows:

1. Perform pattern matching of the form `constructor(subpattern, rest)`, where `subpattern` is a subpattern of the or-pattern, starting from the first such subpattern and proceeding in declarative order.
2. Otherwise pattern matching fails.
```

**Changes to `src/ownership-and-deconstruction.rst`**

Add the following text **before** `15.9:3 Binding are`:
```
* Bindings introduced by an or-pattern are dropped in reverse declaration order, where the declaration order is defined by the first subpattern.
```
Change `15.9:3 Binding are` to:
```
All other bindings are dropped in reverse declaration order.
```
The example you provided should be cleaned up a bit, and should be an explanation similar to the existing example.

PLeVasseur:
Thank you for the extensive feedback @kirtchev-adacore 🥰

I'll review and then update or discuss as needed.

tshepang:
one thing we should decide is if we should add (new) code examples to FLS, and if so, in what cases

kirtchev-adacore:
I think that these new drop semantics are sufficiently complex that they deserves an example. In fact, I had to run the example code in the Rust Playground to ensure I understood the semantics correctly.

PLeVasseur:
> Hi Pete! Here is some feedback, which attempts to make the new text a bit more formal. Note that I am actually introducing more rules to cover the missing semantics for or-patterns.

Thanks again @kirtchev-adacore -- I've made a revision at: https://github.com/rust-lang/fls/pull/631/changes/59e9df0ee3e675943f9c99a9de7d0728b65c6245

> The example you provided should be cleaned up a bit, and should be an explanation similar to the existing example.

Haven't done this one yet. Maybe we can discuss in today's meeting.

src/glossary.rst
```diff
@@ -4713,6 +4713,19 @@ opt-out trait bound
...
 that nullifies an implicitly added :t:`trait bound`.
 
+
+.. _fls_LnPDQW3bnNUw:
+
+or-pattern
+^^^^^^^^^^
+
+:dp:`fls_LnPDQW3bnNUw`
+An :dt:`or pattern` is a :t:`pattern` that matches on one of two or more
```
PLeVasseur: via @kirtchev-adacore -- add the following here:
```
is a :t:`pattern` which or-s two or more :t:`[subpattern]s` using
character 0x7C (vertical line).
```

And then mirror the glossary into the text where this definition is used.

(part of why @tshepang wants the glossary as a source document to go away, but render it instead from the body of the text)

src/patterns.rst
```diff
@@ -65,18 +69,24 @@ of a :t:`pattern-without-range`, except for when the :t:`pattern` is ``&mut``
 :s:`Identifier`. Such a :t:`pattern` is interpreted as a :t:`reference pattern`
 with :t:`keyword` ``mut`` containing an :t:`identifier pattern`.
 
+:dp:`fls_72JHo343O7jp`
+An or-pattern shall not appear in the :t:`pattern-without-alternation` of a
+:t:`closure parameter`, a :t:`function parameter`, or a :t:`let statement`.
```
PLeVasseur: Requires definition of `let binding`.

```suggestion
:t:`closure parameter`, a :t:`function parameter`, or a :t:`let binding`.
```

Thanks to @traviscross, @ehuss for pointing out the difference and how handled in the Reference.

src/patterns.rst
```diff
@@ -65,18 +69,24 @@ of a :t:`pattern-without-range`, except for when the :t:`pattern` is ``&mut``
...
   :t:`unifiable`.
 
+* :dp:`fls_YDVgFaTQwcL8`
+  The set of :t:`[binding]s` introduced by the two :t:`[pattern]s` shall be distinct.
+
 * :dp:`fls_kv533rntni1x`
-  The :t:`[binding]s` of the two :t:`[pattern-without-alternation]s` shall
-  be the same, shall have :t:`[unifiable type]s`, and shall have the same
-  :t:`[binding mode]s`.
+  Any two :t:`[binding]s` with the same name in the two :t:`[pattern]s` shall have
```
PLeVasseur:
```suggestion
  Any two :t:`[binding]s` with the same name in the two :t:`[subpattern]s` shall have
```

src/patterns.rst
```diff
@@ -65,18 +69,24 @@ of a :t:`pattern-without-range`, except for when the :t:`pattern` is ``&mut``
...
+Any two :t:`[subpattern]s` of an :t:`or-pattern` are subject to the following
+restrictions:
 
 * :dp:`fls_rpvdfmy3n05a`
-  The :t:`[type]s` of the two :t:`[pattern-without-alternation]s` shall be
+  The :t:`[type]s` of the two :t:`[subpattern]s` shall be
   :t:`unifiable`.
 
+* :dp:`fls_YDVgFaTQwcL8`
+  The set of :t:`[binding]s` introduced by the two :t:`[pattern]s` shall be distinct.
```
PLeVasseur:
```suggestion
  The set of :t:`[binding]s` introduced by the two :t:`[subpattern]s` shall be the same.
```

src/ownership-and-deconstruction.rst
```diff
@@ -748,3 +753,38 @@ proceeds as follows:
    }
    let c = PrintOnDrop("2");
 
+:dp:`fls_THzA0QFdMMJB`
```
PLeVasseur: Try to rework into step-by-step explanation as in the example under [Drop Order](https://rust-lang.github.io/fls/ownership-and-deconstruction.html#drop-order).

src/patterns.rst
kirtchev-adacore: You will have to define what a "let binding" is. 😁

src/glossary.rst
PLeVasseur: Did this in: https://github.com/rust-lang/fls/pull/631/commits/c7f77177e37bd64a16374a5888fd9571f33ee972

src/patterns.rst
PLeVasseur: Tried my hand at the definition in: https://github.com/rust-lang/fls/pull/631/commits/900088b9d7a408596157bfec2c2911b3a1fbc886

Happy to hear feedback

src/ownership-and-deconstruction.rst
PLeVasseur: Attempted this here: https://github.com/rust-lang/fls/pull/631/commits/aba5ea58b9512f1991d3655613fdaad86e97d157

src/glossary.rst
```diff
@@ -3741,6 +3741,15 @@ tests for a less-than-or-equals relationship.
...
 See :s:`LessThanOrEqualsExpression`.
 
+
+.. _fls_DdZ1ZwjLZTeG:
+
+let binding
+^^^^^^^^^^^
+
+:dp:`fls_sw6HrsxsnG2y`
+A :dt:`let binding` is a :t:`construct` that introduces :t:`[binding]s` by matching a :t:`value` against a :t:`pattern` using the keyword ``let``. :t:`[let binding]s` appear in :t:`[let statement]s`, :t:`[if let expression]s`, :t:`[while let loop expression]s`, and :t:`[let initializer]s`.
```
kirtchev-adacore: A couple of things here:

I do not think that a `let binding` is a construct. The construct (the "syntactic thing") is actually `PatternWithoutAlternation`. "let binding" is a semantic notion for certain `PatternWithoutAlternation`s (those that appear in let statements, if let expressions, etc).

I do not think that `let initializer` should be included in the context list since the syntax of `LetInitializer` does not contain a `PatternWithoutAlternation`.

How about something along the lines of:

A `let binding` is the `binding` introduced by a `let statement`, an `if let expression`, or a `while let loop expression`.

src/glossary.rst
```diff
@@ -4713,6 +4722,18 @@ opt-out trait bound
...
 that nullifies an implicitly added :t:`trait bound`.
 
+
+.. _fls_LnPDQW3bnNUw:
+
+or-pattern
+^^^^^^^^^^
+
+:dp:`fls_LnPDQW3bnNUw`
+An :dt:`or pattern` is a :t:`pattern` that matches on one of two or more :t:`[pattern-without-alternation]s` and or-s two or more :t:`[subpattern]s` using character 0x7C (vertical line).
```
kirtchev-adacore: Maybe

```suggestion
An :dt:`or pattern` is a :t:`pattern` that matches on one of two or more :t:`[pattern-without-alternation]s` and or-s them using character 0x7C (vertical line).
```

I realized that `pattern-without-alternation` and `subpattern` are not interchangeable, as the latter implies a pattern nested within another pattern at any depth. In the case of `or-pattern`s, the `pattern-without-alternation`s are at the same top level depth.

I suspect that we have other such misuses in the Patterns chapter 😢

src/ownership-and-deconstruction.rst
```diff
@@ -695,7 +695,12 @@ When a :t:`drop scope` is left, all :t:`[value]s` associated with that
 :t:`drop scope` are :t:`dropped` as follows:
 
 * :dp:`fls_g07zq3n55094`
-  :t:`[Binding]s` are :t:`dropped` in reverse declaration order.
+  All other :t:`bindings` are :t:`dropped` in reverse declaration order.
+
+* :dp:`fls_W2S2FrkuedYC`
+  :t:`[Binding]s` introduced by an :t:`or-pattern` are dropped in reverse
+  declaration order, where the declaration order is defined by the first
+  :t:`subpattern`.
```
kirtchev-adacore: Maybe

```suggestion
  :t:`pattern-without-alternation`.
```
depending on what we agree above.

src/ownership-and-deconstruction.rst
kirtchev-adacore: It looks good to me. 👍 

What do others think?

src/patterns.rst
```diff
@@ -41,6 +41,9 @@ Patterns
 A :t:`pattern` is a :t:`construct` that matches a :t:`value` which satisfies all
 the criteria of the :t:`pattern`.
 
+:dp:`fls_VQMmveZUfNTn`
+An :t:`or pattern` is a :t:`pattern` that matches on one of two or more :t:`[pattern-without-alternation]s` and or-s two or more :t:`[subpattern]s` using character 0x7C (vertical line).
```
kirtchev-adacore: This line may need to be updated, depending on what we agree for the glossary entry.

src/patterns.rst
```diff
@@ -65,18 +68,24 @@ of a :t:`pattern-without-range`, except for when the :t:`pattern` is ``&mut``
...
+
 :dp:`fls_8luyomzppck`
-Any two :t:`[pattern-without-alternation]s` that are or-ed using character 0x7C
-(vertical line) are subject to the following restrictions:
+Any two :t:`[subpattern]s` of an :t:`or-pattern` are subject to the following
+restrictions:
 
 * :dp:`fls_rpvdfmy3n05a`
-  The :t:`[type]s` of the two :t:`[pattern-without-alternation]s` shall be
+  The :t:`[type]s` of the two :t:`[subpattern]s` shall be
```
kirtchev-adacore: Maybe

```suggestion
  The :t:`[type]s` of the two :t:`[pattern-without-alternation]s` shall be
```
I realize I may have mislead you during the first review. Sorry for that! 😁

src/patterns.rst
```diff
@@ -65,18 +68,24 @@ of a :t:`pattern-without-range`, except for when the :t:`pattern` is ``&mut``
...
   :t:`unifiable`.
 
+* :dp:`fls_YDVgFaTQwcL8`
+  The set of :t:`[binding]s` introduced by the two :t:`[subpattern]s` shall be the same.
```
kirtchev-adacore: Ditto here.

```suggestion
  The set of :t:`[binding]s` introduced by the two :t:`[pattern-without-alternation]s` shall be the same.
```

src/patterns.rst
```diff
@@ -65,18 +68,24 @@ of a :t:`pattern-without-range`, except for when the :t:`pattern` is ``&mut``
...
+
 * :dp:`fls_kv533rntni1x`
-  The :t:`[binding]s` of the two :t:`[pattern-without-alternation]s` shall
-  be the same, shall have :t:`[unifiable type]s`, and shall have the same
-  :t:`[binding mode]s`.
+  Any two :t:`[binding]s` with the same name in the two :t:`[subpattern]s` shall have
```
kirtchev-adacore: Ditto here.

```suggestion
  Any two :t:`[binding]s` with the same name in the two :t:`[pattern-without-alternation]s` shall have
```

src/patterns.rst
```diff
@@ -1296,6 +1305,9 @@ Pattern Matching
 :t:`Pattern matching` that involves a :t:`pattern` and a context :t:`value`
 proceeds as follows:
 
+#. :dp:`fls_tZJgZDWVChJV`
+   If the pattern is an :t:`or-pattern`, then perform or-pattern matching.
```
kirtchev-adacore:
```suggestion
   If the :t:`pattern` is an :t:`or-pattern`, then perform :t:`or-pattern matching`.
```

src/patterns.rst
```diff
@@ -1350,6 +1362,25 @@ proceeds as follows:
...
 
+
+.. _fls_VsBXBj4AqCj1:
+
+Or-pattern Matching
+~~~~~~~~~~~~~~~~~~~
+
+
+:dp:`fls_njpiXkgi8Ryb`
+Or-pattern matching of an :t:`or-pattern` of the form `constructor(or-pattern, rest)`,
```
kirtchev-adacore:
```suggestion
Or-pattern Matching
~~~~~~~~~~~~~~~~~~~

:dp:`fls_njpiXkgi8Ryb`
:dt:`Or-pattern matching` of an :t:`or-pattern` of the form `constructor(or-pattern, rest)`,
```

src/patterns.rst
```diff
@@ -1350,6 +1362,25 @@ proceeds as follows:
...
+where `constructor` is an arbitrary constructor, `or-pattern` is the :t:`or-pattern`,
+and `rest` is optionally a remaining pattern, proceeds as follows:
+
+#. :dp:`fls_nE7qZpHy9TPu`
+   Perform pattern matching of the form `constructor(subpattern, rest)`, where
+   `subpattern` is a :t:`subpattern` of the :t:`or-pattern`, starting from the
+   first such :t:`subpattern` and proceeding in declarative order.
+#. :dp:`fls_P8yB2b5enpw7`
+   Otherwise pattern matching fails.
```
kirtchev-adacore:
```suggestion
#. :dp:`fls_nE7qZpHy9TPu`
   Perform pattern matching of the form `constructor(subpattern, rest)`, where
   `subpattern` is a :t:`subpattern` of the :t:`or-pattern`, starting from the
   first such :t:`subpattern` and proceeding in declarative order.

#. :dp:`fls_P8yB2b5enpw7`
   Otherwise pattern matching fails.
```
(reads a little better at the source level)

kirtchev-adacore:
@ehuss @traviscross Should the term for "or patterns" have a hyphen? Basically "or-pattern" instead of "or pattern".

ehuss:
We use a hyphen for "or-patterns". 

(aside: TC and I spent entirely too much time yesterday discussing if it is `built-in` or `builtin`.)

traviscross:
Agreed.  Added this entry to our WIP style guide:

#### or-patterns

The hyphen is appropriate in the compound term *or-patterns*, even when it acts as a noun, to avoid a miscue.  The word *or* is acting as a noun adjunct but the reader is going to want to see it as a coordinating conjunction.

*And-gate* and *or-gate*, as nouns, are also hyphenated for this reason.

Note that this is a special circumstance. A term such as *while loop* is not hyphenated except when acting as a phrasal adjective preceding a noun, as in "the *while-loop* expression is...".

kirtchev-adacore:
Thanks for the clarification! @PLeVasseur , I think there are a few uses of "or pattern" in the PR.

src/glossary.rst
PLeVasseur: This makes sense, thanks for catching it.

src/patterns.rst
PLeVasseur: Yep, updated.

src/glossary.rst
PLeVasseur: This looks more faithful. Adopted.

src/ownership-and-deconstruction.rst
```diff
@@ -748,3 +753,27 @@ proceeds as follows:
    }
    let c = PrintOnDrop("2");
 
+:dp:`fls_THzA0QFdMMJB`
+When an :t:`or pattern` is used, the drop order of :t:`[binding]s` is determined by the first :t:`pattern-without-alternation`, regardless of which matches at runtime.
```
kirtchev-adacore:
```suggestion
When an :t:`or-pattern` is used, the drop order of :t:`[binding]s` is determined by the first :t:`pattern-without-alternation`, regardless of which matches at runtime.
```
Apparently I missed this during my precious reviews 😢

src/ownership-and-deconstruction.rst
```diff
@@ -748,3 +753,27 @@ proceeds as follows:
...
+Dropping proceeds as follows:
+
+#. :dp:`fls_zxFM7EoE2Xq8`
+   The first :t:`pattern-without-alternation` ``Ok([a, b])`` declares ``a`` before ``b``.
+
+#. :dp:`fls_093YxG6YXQz2`
+   When the first call matches ``Ok([a, b])``, the :t:`[binding]s` are dropped in reverse declaration order: ``b`` then ``a``.
+
+#. :dp:`fls_gNWXh61ZXXt8`
+   When the second call matches ``Err([b, a])``, the drop order remains ``b`` then ``a``, determined by the first :t:`pattern-without-alternation`.
```
kirtchev-adacore:
```suggestion
   When the second call matches ``Err([b, a])``, the drop order remains ``b`` then ``a`` since it is determined by the first :t:`pattern-without-alternation`.
```
I think this is slightly clearer.

src/glossary.rst
```diff
@@ -4713,6 +4722,18 @@ opt-out trait bound
...
 that nullifies an implicitly added :t:`trait bound`.
 
+
+.. _fls_LnPDQW3bnNUw:
+
+or-pattern
+^^^^^^^^^^
+
+:dp:`fls_LnPDQW3bnNUw`
+An :dt:`or-pattern` is a :t:`pattern` that matches on one of two or more :t:`[pattern-without-alternation]s` and or-s them using character 0x7C (vertical line).
```
traviscross: Perhaps it's worth including the literal `|` here somewhere.

src/ownership-and-deconstruction.rst
```diff
@@ -695,7 +695,12 @@ When a :t:`drop scope` is left, all :t:`[value]s` associated with that
 :t:`drop scope` are :t:`dropped` as follows:
 
 * :dp:`fls_g07zq3n55094`
-  :t:`[Binding]s` are :t:`dropped` in reverse declaration order.
+  All other :t:`bindings` are :t:`dropped` in reverse declaration order.
+
+* :dp:`fls_W2S2FrkuedYC`
+  :t:`[Binding]s` introduced by an :t:`or-pattern` are dropped in reverse
+  declaration order, where the declaration order is defined by the first
+  :t:`pattern-without-alternation`.
```
traviscross: It might seem odd to refer to "all *other* bindings" ahead of speaking to what bindings aren't the other ones.  It might be worth reordering this.

src/ownership-and-deconstruction.rst
traviscross: This feels maybe a bit repetitive with what's above.

src/patterns.rst
```diff
@@ -1296,6 +1305,9 @@ Pattern Matching
 :t:`Pattern matching` that involves a :t:`pattern` and a context :t:`value`
 proceeds as follows:
 
+#. :dp:`fls_tZJgZDWVChJV`
+   If the :t:`pattern` is an :t:`or-pattern`, then perform :t:`or-pattern matching`.
+
```
traviscross, line 1313: This duplicates the algorithm that falls immediately below it.  The text that follows is:

> For each `pattern-without-alternation` of the pattern:

That is, it's meaning to loop over all of the subpatterns of an or-pattern.

src/patterns.rst
```diff
@@ -1350,6 +1362,25 @@ proceeds as follows:
...
+and `rest` is optionally a remaining pattern, proceeds as follows:
+
+#. :dp:`fls_nE7qZpHy9TPu`
+   Perform pattern matching of the form `constructor(subpattern, rest)`, where
+   `subpattern` is a :t:`subpattern` of the :t:`or-pattern`, starting from the
+   first such :t:`subpattern` and proceeding in declarative order.
+
+#. :dp:`fls_P8yB2b5enpw7`
+   Otherwise pattern matching fails.
+
```
traviscross: Why does this look outside of the or-pattern by describing it in terms of `constructor(or-pattern, rest)` (rather than just starting at the or-pattern)?  At first, I thought maybe this was trying to follow some existing pattern in the chapter, but I don't see other instances of this.

kirtchev-adacore: I had suggested that text to Pete, which I extracted from https://github.com/rust-lang/reference/blob/master/src/patterns.md#dynamic-semantics.

src/glossary.rst
PLeVasseur: Wrote in: https://github.com/rust-lang/fls/pull/631/commits/fb345071894357537a0a22e1eedbe911775c5abc

src/ownership-and-deconstruction.rst
PLeVasseur: With this bit?

```
* :dp:`fls_W2S2FrkuedYC`
  :t:`[Binding]s` introduced by an :t:`or-pattern` are dropped in reverse
  declaration order, where the declaration order is defined by the first
  :t:`pattern-without-alternation`.
```

It does seem to have heavy overlap with the bit commented on here mentioning runtime as the difference.

src/patterns.rst
PLeVasseur: This is a good catch. Let's discuss.

src/patterns.rst
PLeVasseur: Ah I see. Okay. Let's discuss in the meeting.

src/ownership-and-deconstruction.rst
PLeVasseur: Will take a look at the example see if there's anything needed from this bit, otherwise remove.

src/patterns.rst
PLeVasseur: Pete to look at what Reference does to specify the runtime semantics and matching.

src/patterns.rst
PLeVasseur: Per @traviscross -- this is relevant to a different thing the Reference is specifying. TC said this verbally, but I wasn't able to follow all the details.

src/ownership-and-deconstruction.rst
PLeVasseur: Per meeting discussion this is an ordering issue.

PLeVasseur: Moved ordering around in: https://github.com/rust-lang/fls/pull/631/commits/69adb5caf52266b05b8f9a33fa221d974145b0fd

src/ownership-and-deconstruction.rst
PLeVasseur: I think the example stands on its own and doesn't need this.

Removed in: https://github.com/rust-lang/fls/pull/631/commits/f2b36cf7eb1e810676604b2f4ce67bce9f4b25ef

src/patterns.rst
PLeVasseur: Tried to rework this here: https://github.com/rust-lang/fls/pull/631/commits/0e7edd8eb59bdba536c142f8f6d0e092154f8748

src/patterns.rst
PLeVasseur: Tried to rework this here: https://github.com/rust-lang/fls/pull/631/commits/0e7edd8eb59bdba536c142f8f6d0e092154f8748

kirtchev-adacore: I think we should clean up the Pattern Matching section a bit since some of the Legality Rules content is better moved into Dynamic Semantics. Here is what I propose:

## Pattern Matching

### Legality Rules

Pattern matching is the process of matching a pattern against a value.

An identifier pattern with keyword mut shall require that the context value is a mutable place expression. (This requirement is moved from Identifier Pattern Matching to here)

### Dynamic Semantics

Pattern matching that involves a pattern and a context value proceeds as follows:

1. If the pattern-without-alternation is an identifier pattern, then perform identifier pattern matching. (Note that "For each pattern-without-alternation of the pattern:" has been removed)
2. (and so on)

Pattern matching that involves an or-pattern and a context value proceeds as follows:

For each subpattern of the or-pattern, starting from the first such subpattern and proceeding in declaratove order:
1. Perform pattern matching with the subpattern against the value.
2. If matching succeeds, stop.

(I am not sure what happens when none of the subpatterns are matched. According to the FLS, "The patterns of all match arms taken together shall exhaustively match the subject expression's type." and similar rules, but perhaps there is Undefined Behavior here that we need to mention?)

src/patterns.rst
kirtchev-adacore: Hmm, I wonder whether we need to have this section. Given the suggestion I proposed above, the existing Pattern Matching section models roughly the following functionality (expressed in pseudocodish mumbo-jumbo):
```
pattern_match(pattern) -> bool
    if pattern is identifier_pattern and identifier_pattern_match(pattern), then
        return True
    else if pattern is or_pattern, then
        for subpattern of pattern
            if pattern_match(subpattern), then
                return True
        (undefined behavior here?)
    return False
```
The point is that subpatterns are recursively handled by the Pattern Matching section.

PLeVasseur: The goal was to make it a bit more explicit, the jumping around that we're doing. Maybe wasn't capture clearly.

It could be made flat and within the Pattern Matching section entirely, even for removing the or-patterns heading I think. I was debating doing that in a similar manner to the pseudocode you wrote.

If that looks good, that's the path I'll take, cc @tshepang @traviscross for visibility

kirtchev-adacore: > It could be made flat

I thought about that too. Since we have dedicated subsections for the individual pattern kinds, I figured we can keep the symmetry when discussing Pattern Matching.

I also considered moving the Runtime Semantics of the individual patterns within the subsections of the patterns themselves, but this seemed to break up the semantics of Pattern Matching too much, and required the use of terms and notions what were not yet defined.

PLeVasseur: I tried to make it flat and recurse back into itself here: https://github.com/rust-lang/fls/pull/631/changes/f7f691df8158e278410fecbfac52a3f17ca5096f

Thoughts?

kirtchev-adacore, line 1306: I think you should add
```
:dt:`Pattern matching` is the process of matching a :t:`pattern` against a
:t:`value`.
```
here, and change the entry in the glossary to be a term reference (`:t:`).

kirtchev-adacore: > Thoughts?

LGTM!

PLeVasseur: Yeah, that seems reasonable. Will do.

PLeVasseur: Gave this a go in: https://github.com/rust-lang/fls/pull/631/changes/bda071550fc7caf3fb8c08ccbc24ea23467333bf

PLeVasseur:
Looks like build failed due to Python formatting, so I opened an issue to track this: https://github.com/rust-lang/fls/issues/642

Probably we should have a PR to update the formatting first, before this merges.

cc @tshepang, @kirtchev-adacore, @traviscross

tshepang:
https://github.com/rust-lang/fls/pull/644

tshepang:
will need a rebase to rid of CI fails

PLeVasseur:
Rebased, building, ready for hopefully one last cursory review before we bring it in.

cc @kirtchev-adacore, @tshepang

kirtchev-adacore:
LGTM! There are some outstanding comments from @traviscross , so I will let him resolve those.

traviscross:
Looks good to me.  Thanks @PLeVasseur, @kirtchev-adacore.

PR #636, opened 2025-12-15
repr(packed) can take arguments

tshepang:


src/attributes.rst
```diff
@@ -1543,7 +1544,7 @@ Attribute ``repr``
      | $$usize$$
 
    AlignmentDecrease ::=
-       $$packed$$
+       $$packed$$ $$($$ DecimalLiteral $$)$$?
```
kirtchev-adacore: Unfortunately the grammar is not correct. As currently formulated, only the closing parenthesis is optional. You will need to introduce an intermediate production:
```
$$packed$$ XXX?

XXX ::=
    $$($$ DecimalLiteral $$)$$
```
Unfortunately I cannot come up with a good name for `XXX`. `PackedValue`, `AlignmentValue`, something else.

I am not sure why attribute `repr` lacks more legality rules, but I think we should at least consider adding a few rules for [this text](https://doc.rust-lang.org/nomicon/other-reprs.html) and [this text](https://doc.rust-lang.org/stable/reference/type-layout.html#the-alignment-modifiers).

tshepang: would using a bracket work, like `$$packed$$ ($$($$ DecimalLiteral $$)$$)?`

kirtchev-adacore: I considered that, but I found the mix of grouping parentheses and syntactic parentheses a bit confusing. Note that you may also consider the following refactor:
```
AlignmentDecrease ::=
    packed AlignmentValue?

AlignmentIncrease ::=
   align AlignmentValue

AlignmentValue ::=
    $$($$ DecimalLiteral $$)$$
```

PR #638, opened 2025-12-18
add changelog entries for 1.92 release

tshepang:


src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
    just the language changes that had an impact to the FLS. See the `release
    notes`_ for a full list of changes.
 
+Language changes in Rust 1.92.0
+-------------------------------
+
+- `Document MaybeUninit representation and validity <https://github.com/rust-lang/rust/pull/140463>`_
```
PLeVasseur, line 25: Does that mean we should open an issue to document `MaybeUninit`? Just double-checked and it's in the 2021 edition.

src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
...
+
+  - This is not a language change
+
+- `Allow "&raw [mut | const]" for union field in safe code <https://github.com/rust-lang/rust/pull/141469>`_
```
PLeVasseur, line 29: Does that mean we should open an issue to document `union`s more comprehensively?

src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
...
+
+  - This lifted restriction was not specified in the FLS
+
+- `Prefer item bounds of associated types over where-bounds for auto-traits and Sized <https://github.com/rust-lang/rust/pull/144064>`_
+
+  - This is a fix for a regression that only existed in development versions of Rust (Beta and Nightly)
+
+- `Do not materialize X in "[X; 0]" when X is unsizing a const <https://github.com/rust-lang/rust/pull/145277>`_
```
PLeVasseur, line 37: Should we open an issue to more comprehensively document this?

Maybe speaks to "how exhaustive" the FLS should be vs can be currently.

src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
...
+
+  - This behavior was not documented in the FLS
+
+- `Support combining #[track_caller] and #[no_mangle] (requires every declaration specifying #[track_caller] as well) <https://github.com/rust-lang/rust/pull/145724>`_
```
PLeVasseur, line 41: Should we open an issue to document this? Both `#[no_mangle]` and `#[track_caller]` feature in the FLS.

src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
...
+
+  - This lifted restriction was not documented in the FLS
+
+- `Make never type lints never_type_fallback_flowing_into_unsafe and dependency_on_unit_never_type_fallback deny-by-default <https://github.com/rust-lang/rust/pull/146167>`_
+
+  - Lints are not part of the FLS
+
+- `Allow specifying multiple bounds for same associated item, except in trait objects <https://github.com/rust-lang/rust/pull/146593>`_
```
PLeVasseur, line 49: Should we open an issue for this? The FLS does document these categories of things.

src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
...
+
+  - This lifted restriction was not documented in the FLS
+
+- `Slightly strengthen higher-ranked region handling in coherence <https://github.com/rust-lang/rust/pull/146725>`_
```
PLeVasseur, line 53: Should we open an issue to document this? Again, goes to "what we can document now" vs "what should probably be documented".

PLeVasseur (commented):
Less "comments on @tshepang's lovely work", more "how should we approach those items which need not be documented due to the FLS not including that level of granularity currently?"

src/changelog.rst
PLeVasseur: Per discussion, this is defined in `core`, therefore not part of the documentation in the Reference, either.

src/changelog.rst
PLeVasseur: Per discussion, this did also not imply a Reference update. Bug fix.

src/changelog.rst
PLeVasseur: Per discussion, this did also not imply a Reference update. Bug fix.

src/changelog.rst
PLeVasseur: Per discussion, this did also not imply a Reference update. Bug fix.

src/changelog.rst
PLeVasseur: Per discussion, this did also not imply a Reference update. This is a lifted restriction which does not imply an FLS update.

We could consider making the [Generic Conformance section](https://rust-lang.github.io/fls/generics.html#generic-conformance) more detailed.

src/changelog.rst
PLeVasseur: Per discussion, this did also not imply a Reference update. This is a lifted restriction and would require more granular handling to be relevant. Seems outside the scope of FLS currently.

PR #652, opened 2026-01-29
add missing changelog entries for #631

tshepang:


src/changelog.rst
```diff
@@ -100,6 +100,39 @@ Language changes in Rust 1.91.0
...
+    - :p:`fls_dhfIPP4yR3Tt`
+    - :p:`fls_zxFM7EoE2Xq8`
+    - :p:`fls_093YxG6YXQz2`
+    - :p:`fls_gNWXh61ZXXt8`
+    - :p:`fls_VQMmveZUfNTn`
+    - :p:`fls_72JHo343O7jp`
+    - :p:`fls_YDVgFaTQwcL8`
+    - :p:`fls_zv73CR8rplIa`
+    - :p:`fls_tZJgZDWVChJV`
+
```
PLeVasseur, line 134: In the "New paragraphs" list for this item, I think we're missing a few IDs from PR #631's spec changes:

```suggestion
    - :p:`fls_LnPDQW3bnNUw`
    - :p:`fls_sw6HrsxsnG2y`
    - :p:`fls_urIJ5JNHLhm6`

```

src/changelog.rst
```diff
@@ -100,6 +100,39 @@ Language changes in Rust 1.91.0
...
+    - :p:`fls_am5h8r887bz5`
+    - :p:`fls_eppmiloh7bgg`
+    - :p:`fls_gwc08xayno7q`
+    - :p:`fls_19iygu12s315`
+    - :p:`fls_r307spfk6cs9`
+    - :p:`fls_qhdofvbso3gl`
+    - :p:`fls_drb114dtvlpt`
+    - :p:`fls_uxysntb3u03j`
+    - :p:`fls_vstdqifqipbh`
+
```
PLeVasseur, line 120: In the "Changed paragraphs" list, two IDs from PR 631's diff look missing. One is a list-structure shift (`fls_wh201rmh6u6d`), the other switches `:dt:` to `:t:` (`fls_y3oputy9e0sz`).

```suggestion
    - :p:`fls_wh201rmh6u6d`
    - :p:`fls_y3oputy9e0sz`

```

tshepang: skipped those because I see them not as semantic changes

src/changelog.rst
tshepang: skipped those since they are glossary entries, but am not sure I should have... what say @kirtchev-adacore

tshepang: how I done it in the past is I would define the term inline, not in glossary, and it's then that I would add it to changelog

kirtchev-adacore: I think we should treat new paragraphs in the glossary as "regular" new paragraphs, and document them in the changelog.

tshepang: as agreed in the FLS meet, I replicated the glossary entry inline so that we can put it in the changelog, since the glossary is informational

also
- `fls_sw6HrsxsnG2y` already has an entry inline
- `fls_urIJ5JNHLhm6` is a glossary-only reference, pointing to the section where the glossary entry is introduced inline

https://github.com/rust-lang/fls/pull/652/commits/f667314360e4747bfeedbf0ebf48286f3a2da916

PLeVasseur (approved):
Thanks @tshepang!

Issue #647, opened 2026-01-26
[Change]: [1.93] Stabilize asm_cfg


PLeVasseur:
### Change description

Stabilize asm_cfg

### Area

inline assembly

### Category

New content

### Change description (extended)

[Stabilize asm_cfg](https://github.com/rust-lang/rust/pull/147736)

PR #653, opened 2026-01-31
Document asm_cfg inline assembly attributes

PLeVasseur:
## Summary
- document `asm_cfg` attribute support on inline assembly templates/operands and update the inline-assembly grammar
- add attribute legality rules and an example under the asm macros section
- record syntax/paragraph updates in the Rust 1.93.0 changelog entry
- Closes rust-lang/fls#647

## Reference alignment
- Reference PR: https://github.com/rust-lang/reference/pull/2063
- Reference sections: https://doc.rust-lang.org/reference/inline-assembly.html#syntax, https://doc.rust-lang.org/reference/inline-assembly.html#attributes
- Grammar: Reference introduces `AsmAttrFormatString`/`AsmAttrOperand`; FLS introduces `AssemblyTemplate`/`AssemblyAttribute*` wrappers to match, splitting operands into `RegisterArgument`, `AbiClobber`, and `AssemblyOption`
- Legality: `cfg`/`cfg_attr` support and the template-first rule match the Reference; FLS makes cfg-false argument behavior explicit to align with conditional compilation semantics
- Reference note about rustc attribute handling is omitted because it is implementation detail without semantic effect

## Testing
- ./make.py

### DIFF
--- a/src/inline-assembly.rst
+++ b/src/inline-assembly.rst
@@ -1659,11 +1659,8 @@ An :t:`inline assembly argument` is either an :t:`assembly instruction`, a :t:`r
 :dp:`fls_nLBhw2w6uznH`
 Only :t:`attribute` :c:`cfg` and :t:`attribute` :c:`cfg_attr` shall decorate :t:`[inline assembly arguments]s` in :s:`AsmArguments` and :s:`GlobalAsmArguments`.
 
-:dp:`fls_xzDPz2zfRfoI`
-If a :t:`assembly instruction`, :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` in :s:`AsmArguments` or :s:`GlobalAsmArguments` is subject to :t:`attribute` :c:`cfg` or :t:`attribute` :c:`cfg_attr` where the related :t:`configuration predicate` evaluates to ``false``, then that construct is not considered part of the related :s:`AsmArguments` or :s:`GlobalAsmArguments`, consistent with :t:`conditional compilation`.
-
 :dp:`fls_cTEiqjf6haEg`
-It is a static error for a :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` to appear before the first :t:`assembly instruction`, including when that construct is subject to :t:`attribute` :c:`cfg` or :t:`attribute` :c:`cfg_attr` where the related :t:`configuration predicate` evaluates to ``false``.
+It is a static error for a :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` to appear before the first :t:`assembly instruction`, including when that :t:`inline assembly argument` is subject to :t:`attribute` :c:`cfg` or :t:`attribute` :c:`cfg_attr` where the related :t:`configuration predicate` evaluates to ``false``.
 
 :dp:`fls_1ikzov7cxic1`
 When invoking :t:`macro` :std:`core::arch::asm`, the :s:`DelimitedTokenTree` of
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -44,7 +44,6 @@ Language changes in Rust 1.93.0
 
     - :p:`fls_tKj18krZ1pSt`
     - :p:`fls_nLBhw2w6uznH`
-    - :p:`fls_xzDPz2zfRfoI`
     - :p:`fls_cTEiqjf6haEg`
 
 - `During const-evaluation, support copying pointers byte-by-byte <https://github.com/rust-lang/rust/pull/148259>`_


PLeVasseur:
Hey @tshepang, @kirtchev-adacore -- could this be reviewed?

src/inline-assembly.rst
```diff
@@ -1641,6 +1653,18 @@ Macros: asm, global_asm, and naked_asm
 :std:`core::arch::global_asm`, and
 :std:`core::arch::naked_asm`.
 
+:dp:`fls_m0SBtonaNppV`
+The :t:`[assembly instruction]s`, :t:`[register argument]s`, :t:`[ABI clobber]s`, and :t:`[assembly option]s` in :s:`AsmArguments` and :s:`GlobalAsmArguments` may be preceded by :t:`outer attribute` instances.
```
kirtchev-adacore: There is no need to state the optionality of outer attributes because this is apparent from the syntax. If the optionality was dependent on some semantic condition, then absolutely, but this is not the case here.

src/inline-assembly.rst
```diff
@@ -1641,6 +1653,18 @@ Macros: asm, global_asm, and naked_asm
...
+
+:dp:`fls_nLBhw2w6uznH`
+Only the :t:`attribute` :c:`cfg` and the :t:`attribute` :c:`cfg_attr` are accepted on inline assembly arguments. All other attributes are rejected.
```
kirtchev-adacore: A couple of things here:

1. What are "inline assembly arguments"? This term is not defined. There should be no implied terms nor implied semantics in the FLS, everything must be unambiguously spelled out.
2. I suspect that this paragraph must be expressed as a rule since it looks like a requirement aimed at the programmer. How about

Only the following outer attributes shall decorate an inline assembly argument:
- Attribute cfg
- Attribute cfg_attr

We have used the verb "decorate" with attributes before, see 13.1 Attribute Properties. I am not sure we have been consistent with the usage though... 😆

src/inline-assembly.rst
```diff
@@ -1641,6 +1653,18 @@ Macros: asm, global_asm, and naked_asm
...
+
+:dp:`fls_xzDPz2zfRfoI`
+If a :t:`assembly instruction`, :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` is annotated with :c:`cfg` or :c:`cfg_attr` and the related :t:`configuration predicate` evaluates to ``false``, the annotated argument is not considered part of the related macro invocation, consistent with :t:`conditional compilation`.
+
+:dp:`fls_cTEiqjf6haEg`
+It is a static error for a :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` to appear before the first :t:`assembly instruction`, even if the argument is ignored by :t:`conditional compilation`.
```
kirtchev-adacore: Note that only "register argument" has the word "argument" in its term definition. So you definitely need to define  what "inline assembly arguments" are.

src/inline-assembly.rst
```diff
@@ -1641,6 +1653,18 @@ Macros: asm, global_asm, and naked_asm
...
 :std:`core::arch::naked_asm`.
 
+:dp:`fls_m0SBtonaNppV`
+The :t:`[assembly instruction]s`, :t:`[register argument]s`, :t:`[ABI clobber]s`, and :t:`[assembly option]s` in :s:`AsmArguments` and :s:`GlobalAsmArguments` may be preceded by :t:`outer attribute` instances.
+
+:dp:`fls_nLBhw2w6uznH`
+Only the :t:`attribute` :c:`cfg` and the :t:`attribute` :c:`cfg_attr` are accepted on inline assembly arguments. All other attributes are rejected.
+
+:dp:`fls_xzDPz2zfRfoI`
+If a :t:`assembly instruction`, :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` is annotated with :c:`cfg` or :c:`cfg_attr` and the related :t:`configuration predicate` evaluates to ``false``, the annotated argument is not considered part of the related macro invocation, consistent with :t:`conditional compilation`.
```
kirtchev-adacore:
```suggestion
If an :t:`assembly instruction`, :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` is decorated with :t:`outer attribute` :c:`cfg` or :c:`cfg_attr` and the related :t:`configuration predicate` evaluates to ``false``, then the decorated inline assembly argument is not considered part of the related macro invocation.
```
I am not sure what "considered part of the related macro invocation" means. Does it mean that the argument is removed at the point of macro invocation?

src/inline-assembly.rst
```diff
@@ -1641,6 +1653,15 @@ Macros: asm, global_asm, and naked_asm
 :std:`core::arch::global_asm`, and
 :std:`core::arch::naked_asm`.
 
+:dp:`fls_nLBhw2w6uznH`
```
kirtchev-adacore: Pete,

How about adding a paragraph before this one which defines "inline assembly argument"? Something along the lines of
```
An :t:`inline assembly argument` is either an :t:`assembly instruction`, a :t:`register argument`, an :t:`ABI clobber`, or an :t:`assembly option`.
```
(with a `:dt:inline assembly argument` in the Glossary)

src/inline-assembly.rst
```diff
@@ -1641,6 +1653,15 @@ Macros: asm, global_asm, and naked_asm
...
+Only :t:`attribute` :c:`cfg` and :t:`attribute` :c:`cfg_attr` shall decorate :t:`[assembly instruction]s`, :t:`[register argument]s`, :t:`[ABI clobber]s`, and :t:`[assembly option]s` in :s:`AsmArguments` and :s:`GlobalAsmArguments`; all other :t:`[attribute]s` shall be rejected.
```
kirtchev-adacore:
```suggestion
Only :t:`attribute` :c:`cfg` and :t:`attribute` :c:`cfg_attr` shall decorate :t:`[inline assembly arguments]s in :s:`AsmArguments` and :s:`GlobalAsmArguments`.
```
Note that "all other attributes shall be rejected" does not make sense as a rule because it implies that the programmer shall be rejecting the attributes, not the confirming tool. Normative "shall" and "must" are for the programmer to obey, facts are what a conforming tool guarantees.

src/inline-assembly.rst
```diff
@@ -1641,6 +1653,15 @@ Macros: asm, global_asm, and naked_asm
...
+
+:dp:`fls_xzDPz2zfRfoI`
+If a :t:`assembly instruction`, :t:`register argument`, :t:`ABI clobber`, or :t:`assembly option` in :s:`AsmArguments` or :s:`GlobalAsmArguments` is subject to :t:`attribute` :c:`cfg` or :t:`attribute` :c:`cfg_attr` where the related :t:`configuration predicate` evaluates to ``false``, then that construct is not considered part of the related :s:`AsmArguments` or :s:`GlobalAsmArguments`, consistent with :t:`conditional compilation`.
```
kirtchev-adacore:
```suggestion
If an :t:`inline assembly argument` in :s:`AsmArguments` or :s:`GlobalAsmArguments` is subject to :t:`attribute` :c:`cfg` or :t:`attribute` :c:`cfg_attr` where the related :t:`configuration predicate` evaluates to ``false``, then the :t:`inline assembly argument` is not considered part of the related :s:`AsmArguments` or :s:`GlobalAsmArguments`, consistent with :t:`conditional compilation`.
```
I wonder whether this paragraph is even needed because it is subsumed by

18.6:3 A construct subject to attribute cfg where the related configuration predicate evaluates to false is not considered part of a Rust program.

src/inline-assembly.rst
PLeVasseur: Added a local definition before the legality rule and a matching glossary entry for `:t:`inline assembly argument``.

The new definition is:
`An :t:`inline assembly argument` is either an :t:`assembly instruction`, a :t:`register argument`, an :t:`ABI clobber`, or an :t:`assembly option`.`

This is now present in `src/inline-assembly.rst` and `src/glossary.rst`, and the changelog records the new glossary term.

src/inline-assembly.rst
PLeVasseur: I ended up removing the standalone paragraph.

The generic conditional-compilation rule in 18.6 already covers the cfg-false case, so keeping a separate paragraph here seemed redundant. I kept the asm-specific template-first rule and tightened it so the cfg-disabled operand case is still covered locally where that unique behavior matters.

src/inline-assembly.rst
PLeVasseur: Reworked this into a pure legality rule using the newly defined `inline assembly argument` term:
`Only :t:`attribute` :c:`cfg` and :t:`attribute` :c:`cfg_attr` shall decorate :t:`[inline assembly arguments]s` in `:s:`AsmArguments`` and `:s:`GlobalAsmArguments``.`

I also removed the `all other attributes shall be rejected` clause so the paragraph no longer mixes a programmer-facing rule with tool behavior.

Issue #674, opened 2026-03-10
[Change]: [1.94.0] Avoid incorrect lifetime errors for closures


PLeVasseur:
### Change description

Avoid incorrect lifetime errors for closures

### Area

closures

### Category

New content

### Change description (extended)

Link to compiler PR:
[Avoid incorrect lifetime errors for closures](https://github.com/rust-lang/rust/pull/148329)

kirtchev-adacore:
@PLeVasseur I looked over the original compiler PR discussion, and I did not find a Reference PR. I also skimmed over the compiler changes, and they look like an enhancement/bug fix to me.

@ehuss @traviscross Do you recall whether the compiler PR had any language impact?

ehuss:
We generally don't have borrowck documented in the reference, so that's why it didn't show up there.

It has a language impact insofar that it changed what is accepted in borrowck for closures.

kirtchev-adacore:
@ehuss Thanks for the info!

This is a no-op for the FLS then.

PR #679, opened 2026-03-11
[Change]: [1.94.0] Avoid incorrect lifetime errors for closures

kirtchev-adacore:
This PR updates the changelog of the FLS to indicate that the related rust-lang PR does not fall within the scope of the FLS.

Closes: https://github.com/rust-lang/fls/issues/674


tshepang:
could the wording be more like "exact mechanics of borrow checker are outside the scope of the FLS", which would be a more accurate statement

kirtchev-adacore:
I like it! Updated.

Issue #680, opened 2026-03-18
steps to get to generating glossary automatically


tshepang:
- [x] Remove section and paragraph ids from glossary (#681)
  note: this could come later, but I wanted to see if it could easily be done (then went ahead and doned it)
- [ ] Ensure main text has all glossary terms
- [ ] Ensure main text defines all glossary terms
- [ ] Ensure the duplicate definitions that are in both glossary and main text have matching text
  - If text does not match, choose one that is more accurate
- [ ] Ensure `:dt:` entries are in main text and that there are none in glossary
  - This will mean clicking on the term in a glossary will take one to the main text
- [ ] Remove the "See SomeTerm" paragraphs in the glossary, because above step makes them redundant
- [ ] Generate the glossary from reading all `:dt:` entries from main text, and ensure it matches the old glossary

_Originally posted by @tshepang in https://github.com/rust-lang/fls/issues/656#issuecomment-3876000720_

PLeVasseur:
Nice, good to see this happening! I left a [comment](https://github.com/rust-lang/fls/pull/681) over on the PR.

kirtchev-adacore:
Note: the proposed work plan will need to handle https://github.com/rust-lang/fls/issues/682.

PR #681, opened 2026-03-18
remove needless ids (from glossary)

tshepang:
The glossary is an informational document, and there is no need for it to have section and paragraph ids. These ids are more useful for the normative text, which needs to be referenced by tests and therefore needs to be stable even if names of terms change.

This change is in preparation for automatically generating glossary chapter, instead of maintaining it manually, which risks resulting in mismatches with main text. To make our lives more easy, we have decided to remove these ids, because we would otherwise have to keep track of them and keep them stable, as well as generate fresh ones as needed.

See rust-lang/fls#680.

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -6,7863 +6,5199 @@
 
 .. informational-page::
 
-.. _fls_bc2qwbfibrcs:
-
 Glossary
 ========
-
-.. _fls_m98yg554tj9s:
 
 ABI
 ^^^
 
-:dp:`fls_4ko8qcah0f9k`
 For :dt:`ABI`, see :t:`Application Binary Interface`.
-
-.. _fls_zOdDwoObYHC0:
 
 ABI clobber
 ^^^^^^^^^^^
 
-:dp:`fls_OVX4RFcWKfP9`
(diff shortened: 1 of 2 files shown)


src/glossary.rst
```diff
@@ -6,7863 +6,5199 @@
...
 :dt:`Private visibility` is a kind of :t:`visibility` that allows a :t:`name`
 to be referred to only by the current :t:`module` of the :t:`entity`, and its
 descendant :t:`[module]s`.
 
-.. _fls_kCA6SW8bUq5x:
-
 proc-macro crate
 ^^^^^^^^^^^^^^^^
 
-.. _fls_AjjdLZWiL9Tq:
```
tshepang, line 5149: extra double section id because why not

PLeVasseur:
Howdy @tshepang -- have we talked with Ferrous Systems folks about how this may or may not impact the ability for the assessor to evaluate the FLS? I suppose this information is kind of "under the hood" detail that's not immediately visible, so I wonder.

tshepang:
one only needs to explain to the assessor what is done and why (if they ask), so am not concerned there, especially that we are not changing anything visible

PR #686, opened 2026-03-25
Add dev guide

### LINKED ISSUE #685: Create FLS dev-guide
As discussed during the 2026-03-20 FLS meeting, the FLS should have a dev guide. The purpose of the dev guide would be to explain various formats, procedures, and processes associated with the FLS.

kirtchev-adacore:
This PR adds a new appending to the FLS - the dev guide.

Closes: https://github.com/rust-lang/fls/issues/685

### DIFF
--- a/src/dev-guide.rst
+++ b/src/dev-guide.rst
@@ -1,0 +1,87 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Rust Project Contributors
+
+.. default-domain:: spec
+
+.. informational-page::
+
+Developer Guide
+===============
+
+This document outlines the various formats, processes, and procedures associated with the maintenance of the FLS.
+
+Changelog maintenance
+---------------------
+
+The Changelog is located in ``src/changelog.rst``. It should be updated using one of the sentence patterns outlined below. Note that this is not an exhaustive list, and special cases would need to use their own wording.
+
+No change
+~~~~~~~~~
+
+When an entry in the Rust Release Notes does not affect the FLS, provide the reason as to why this is the case. Below are a few examples of such justifications::
+
+    - Lints are outside the scope of the FLS
+    - Target XYZ is outside the scope of the FLS
+    - The exact mechanics of the borrow checker are outside the scope of the FLS
+    - The restriction was not specified in the FLS
+    - Configuration options are environment-specific and not exhaustive
+
+Glossary entries
+~~~~~~~~~~~~~~~~
+
+When working with glossary entries, use the following sentence pattern for a single glossary entry::
+
+    <Action> glossary entry: :t:`term`
+
+Use the following sentence pattern for multiple paragraphs::
+
+    <Action> glossary entries:
+    - :t:`term`
+    - :t:`term`
+
+``<Action>`` must denote either ``Changed``, ``New``, or ``Removed``.
+
+Paragraphs
+~~~~~~~~~~
+
+When working with paragraphs, use the following sentence pattern for a single paragraph::
+
+    <Action> paragraph: :p:`fls_paragraph_id`
+
+Use the following sentence pattern for multiple paragraphs::
+
+    <Action> paragraphs:
+    - :p:`fls_paragraph_id`
+    - :p:`fls_paragraph_id`
+
+``<Action>`` must denote either ``Changed``, ``Moved``, ``New``, or ``Removed``.
+
+Sections
+~~~~~~~~
+
+When working with sections, use the following sentence pattern for a single section::
+
+    <Action> section :ref:`fls_section_id`
+
+Use the following sentence pattern for multiple sections::
+
+    <Action> sections:
+    - :ref:`fls_section_id`
+    - :ref:`fls_section_id`
+
+``<Action>`` must denote either ``Moved``, ``New``, or ``Removed``.
+
+Syntax
+~~~~~~
+
+When working with syntax, use the following sentence pattern for a single syntax category::
+
+    <Action> syntax: :s:`syntax_category`
+
+Use the following sentence pattern for multiple syntax categories::
+
+    <Action> syntax:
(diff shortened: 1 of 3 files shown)


src/dev-guide.rst
```diff
@@ -0,0 +1,90 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
+   SPDX-FileCopyrightText: The Rust Project Contributors
+
+.. default-domain:: spec
+
+.. informational-page::
+
+Developer's Guide
```
tshepang: maybe we call it "Developer Guide", so as to match what Reference calls [its equivalent](https://rust-lang.github.io/reference/dev-guide) (but feel free to ignore this since it's so nit)

src/dev-guide.rst
```diff
@@ -0,0 +1,90 @@
...
+=================
+
+This document outlines the various formats, processes, and procedures associated with the maintenance of the FLS.
+
+Changelog maintenance
+---------------------
+
+The Changelog is located in ``src/changelog.rst``.
+It should be updated using one of the sentence patterns outlined below.
+Note that this is not an exhaustive list, and special cases would need to use their own wording.
```
tshepang: to follow the [Line length guideline](https://github.com/rust-lang/fls/blob/main/CONTRIBUTING.rst#line-length)
```suggestion
The Changelog is located in ``src/changelog.rst``. It should be updated using one of the sentence patterns outlined below. Note that this is not an exhaustive list, and special cases would need to use their own wording.
```

src/dev-guide.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Glossary entries
+~~~~~~~~~~~~~~~~
+
+When working with glossary entries, use the following sintence pattern for a single glossary entry::
```
tshepang:
```suggestion
When working with glossary entries, use the following sentence pattern for a single glossary entry::
```

src/dev-guide.rst
```diff
@@ -0,0 +1,90 @@
...
+This document outlines the various formats, processes, and procedures associated with the maintenance of the FLS.
+
+Changelog maintenance
+---------------------
+
+The Changelog is located in ``src/changelog.rst``.
+It should be updated using one of the sentence patterns outlined below.
+Note that this is not an exhaustive list, and special cases would need to use their own wording.
+
+Glossary entries
```
tshepang, line 29: not sure we want to document glossary entries since they are informational content (and we are planning to have them be mere duplicates of main content anyways)

src/dev-guide.rst
```diff
@@ -0,0 +1,90 @@
...
+No change
+~~~~~~~~~
+
+Use the following sentence patterns when a PR included in a release does not require a change in the FLS::
+
+    No change: <Reason> is outside the scope of the FLS
+
+or::
+
+    No change: <Reasons> are outside the scope of the FLS
```
tshepang: I would instead use a different example here instead of only add a plural version of the previous example. Maybe you can even list a number of other patterns:

- "No change: this bug was not documented in FLS"
- "No change: this lifted restriction was not specified in the FLS"
- "No change: This behavior was not documented in the FLS"

Also, I do wonder if it's important to have "No change", as the text that follows it should make it clear why the change is not getting documented.

kirtchev-adacore: How about just shrinking this to
```
No change: <Reason>
```
?

I can go either way with keeping/removing the `No change` part. I think we should have a brief discussion about it.

src/dev-guide.rst
kirtchev-adacore: Fixed.

src/dev-guide.rst
kirtchev-adacore: Fixed. I had forgotten about those...

src/dev-guide.rst
kirtchev-adacore: Fixed.

src/dev-guide.rst
kirtchev-adacore: I think we should keep this part in for now. Once the glossary generation machinery is in, we will remove this part.

src/dev-guide.rst
tshepang: yeah, that works... we can maybe try keep the consistency in reviews, so we don't have too many variations for the same information

PLeVasseur: Yeah, let's discuss a bit in the meeting to align.

PLeVasseur (commented):
Thanks for pulling this together @kirtchev-adacore!

I think it looks good overall.

Let's chat a bit about it in our meeting today.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/686/0ff3456e02d31d5ddf9f01b4a0339fc1deadb7b9..ed99280b57b02d43f6358da9f2b21466009bb838)*

src/dev-guide.rst
```diff
@@ -0,0 +1,88 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
```
tshepang:
```suggestion
```

tshepang (commented):
just 1 remaining thing I noticed only now, otherwise good to approve

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/686/0ff3456e02d31d5ddf9f01b4a0339fc1deadb7b9..413ff5387643b2bf1838b603d44f3909e0f99d46)*

src/dev-guide.rst
kirtchev-adacore: The decision we took during the meeting has been incorporated into the PR.

src/dev-guide.rst
kirtchev-adacore: Good catch! Fixed.

tshepang (approved):

PR #689, opened 2026-04-04
ensure all terms in glossary also exist in main text

tshepang:
*It will be more easy to review commits individually.*

This is part of https://github.com/rust-lang/fls/issues/680, and a number of fixes and improvements (I think) were made along the way.

The one commit am not so sure of is bd6ab3fea72f789b5e8ff11161004fe1bbd2aa17... I can imagine doing it differently, but am not sure it that will make the diff smaller.

### DIFF
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -30,6 +30,7 @@ FLS maintenance
   - :p:`fls_twC8XhwbI60x`
   - :p:`fls_9bp4cKSfW9A0`
   - :p:`fls_Odh6Z7ivD1U4`
+  - :p:`fls_FG0EmVgrDUDq`
 
 - Paragraphs changed to avoid referencing indirect terms:
 
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -57,6 +57,10 @@ FLS maintenance
   - :p:`fls_9d970yfwmj2d`
   - :p:`fls_p9687v3xckps`
   - :p:`fls_8j408kckzzud`
+  - :p:`fls_mdejdc0nNgyO`
+  - :p:`fls_WUCvCauUCmsD`
+  - :p:`fls_5SVthCNZBcVS`
+  - :p:`fls_SonodAQVL5J4`
 
 Language changes in Rust 1.94.0
 -------------------------------
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -29,6 +29,32 @@ FLS maintenance
   - :p:`fls_hyi2jnp38v1n`
   - :p:`fls_twC8XhwbI60x`
 
+- Paragraphs changed to avoid referencing indirect terms:
+
+  - :p:`fls_xmgcdw9yhb55`
+  - :p:`fls_yeh6mvyvb4dp`
+  - :p:`fls_657knnsobdyu`
+  - :p:`fls_m942dwwmr2cl`
+  - :p:`fls_np33oqrz33mp`
+  - :p:`fls_atdpr8be2o2r`
+  - :p:`fls_fbgwb3pdfgz`
+  - :p:`fls_8tbxq95x06yt`
+  - :p:`fls_9oy9zo3x3fy3`
+  - :p:`fls_pdgj2xekdead`
+  - :p:`fls_4uoi6k8r7mvc`
+  - :p:`fls_fjaz4m90cagr`
+  - :p:`fls_eesn9kuylim`
+  - :p:`fls_ijfmnnrdlu8n`
+  - :p:`fls_6x7j9x354pkb`
+  - :p:`fls_h2cpbz2t74hy`
+  - :p:`fls_whj50spxz3bh`
+  - :p:`fls_d1cxq1zbt5fq`
+  - :p:`fls_48i245an2449`
+  - :p:`fls_69wr03rt0ali`
+  - :p:`fls_9d970yfwmj2d`
+  - :p:`fls_p9687v3xckps`
+  - :p:`fls_8j408kckzzud`
+
 Language changes in Rust 1.94.0
 -------------------------------
 
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -19,6 +19,16 @@ with the change that has been applied due to it.
    just the language changes that had an impact to the FLS. See the `release
    notes`_ for a full list of changes.
 
+FLS maintenance
+---------------
+
+- Paragraphs added or changed to ensure all glossary terms are also found in the main text:
+
+  - :p:`fls_BBz4Iy7AeaU4`
+  - :p:`fls_cxNAtLSLjqbQ`
+  - :p:`fls_hyi2jnp38v1n`
+  - :p:`fls_twC8XhwbI60x`
+
 Language changes in Rust 1.94.0
 -------------------------------
 
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -28,6 +28,8 @@ FLS maintenance
   - :p:`fls_cxNAtLSLjqbQ`
   - :p:`fls_hyi2jnp38v1n`
   - :p:`fls_twC8XhwbI60x`
+  - :p:`fls_9bp4cKSfW9A0`
+  - :p:`fls_Odh6Z7ivD1U4`
 
 - Paragraphs changed to avoid referencing indirect terms:
 
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -29,6 +29,32 @@ FLS maintenance
   - :p:`fls_hyi2jnp38v1n`
   - :p:`fls_twC8XhwbI60x`
 
+- Paragraphs changed to avoid referencing indirect terms:
+
+  - :p:`fls_xmgcdw9yhb55`
+  - :p:`fls_yeh6mvyvb4dp`
+  - :p:`fls_657knnsobdyu`
+  - :p:`fls_m942dwwmr2cl`
+  - :p:`fls_np33oqrz33mp`
+  - :p:`fls_atdpr8be2o2r`
+  - :p:`fls_fbgwb3pdfgz`
+  - :p:`fls_8tbxq95x06yt`
+  - :p:`fls_9oy9zo3x3fy3`
+  - :p:`fls_pdgj2xekdead`
+  - :p:`fls_4uoi6k8r7mvc`
+  - :p:`fls_fjaz4m90cagr`
+  - :p:`fls_eesn9kuylim`
+  - :p:`fls_ijfmnnrdlu8n`
+  - :p:`fls_6x7j9x354pkb`
+  - :p:`fls_h2cpbz2t74hy`
+  - :p:`fls_whj50spxz3bh`
+  - :p:`fls_d1cxq1zbt5fq`
+  - :p:`fls_48i245an2449`
+  - :p:`fls_69wr03rt0ali`
+  - :p:`fls_9d970yfwmj2d`
+  - :p:`fls_p9687v3xckps`
+  - :p:`fls_8j408kckzzud`
+
 - Remove the term "indirection", and associated derivatives
 
   Changed glossary entries:
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -57,6 +57,10 @@ FLS maintenance
   - :p:`fls_9d970yfwmj2d`
   - :p:`fls_p9687v3xckps`
   - :p:`fls_8j408kckzzud`
+  - :p:`fls_mdejdc0nNgyO`
+  - :p:`fls_WUCvCauUCmsD`
+  - :p:`fls_5SVthCNZBcVS`
+  - :p:`fls_SonodAQVL5J4`
 
 - Remove the term "indirection", and associated derivatives
 
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -22,6 +22,13 @@ with the change that has been applied due to it.
 FLS maintenance
 ---------------
 
+- Paragraphs added or changed to ensure all glossary terms are also found in the main text:
+
+  - :p:`fls_BBz4Iy7AeaU4`
+  - :p:`fls_cxNAtLSLjqbQ`
+  - :p:`fls_hyi2jnp38v1n`
+  - :p:`fls_twC8XhwbI60x`
+
 - Remove the term "indirection", and associated derivatives
 
   Changed glossary entries:
--- a/src/expressions.rst
+++ b/src/expressions.rst
@@ -2606,60 +2606,60 @@ is used in the calculation of a :t:`compound assignment expression`.
 An :t:`assigned operand` shall denote a :t:`mutable assignee expression`.
 
 :dp:`fls_xmgcdw9yhb55`
-The :t:`type` of a :t:`compound assignment` is the :t:`unit type`.
+The :t:`type` of a :t:`compound assignment expression` is the :t:`unit type`.
 
 :dp:`fls_yeh6mvyvb4dp`
-The :t:`value` of a :t:`compound assignment` is the :t:`unit value`.
+The :t:`value` of a :t:`compound assignment expression` is the :t:`unit value`.
 
 :dp:`fls_657knnsobdyu`
-The :t:`type` of the :t:`assigned operand` of an :t:`addition assignment` shall
+The :t:`type` of the :t:`assigned operand` of an :t:`addition assignment expression` shall
 implement the :std:`core::ops::AddAssign` trait where the type of the right
 operand is the trait implementation type parameter.
 
 :dp:`fls_m942dwwmr2cl`
-The :t:`type` of the :t:`assigned operand` of a :t:`bit and assignment` shall
+The :t:`type` of the :t:`assigned operand` of a :t:`bit and assignment expression` shall
 implement the :std:`core::ops::BitAndAssign` :t:`trait` where the :t:`type` of
 the :t:`modifying operand` is the :t:`trait implementation` :t:`type parameter`.
 
 :dp:`fls_np33oqrz33mp`
-The :t:`type` of the :t:`assigned operand` of a :t:`bit or assignment` shall
+The :t:`type` of the :t:`assigned operand` of a :t:`bit or assignment expression` shall
 implement the :std:`core::ops::BitOrAssign` :t:`trait` where the :t:`type` of
 the :t:`modifying operand` is the :t:`trait implementation` :t:`type parameter`.
 
 :dp:`fls_atdpr8be2o2r`
-The :t:`type` of the :t:`assigned operand` of a :t:`bit xor assignment` shall
+The :t:`type` of the :t:`assigned operand` of a :t:`bit xor assignment expression` shall
 implement the :std:`core::ops::BitXorAssign` :t:`trait` where the :t:`type` of
 the :t:`modifying operand` is the :t:`trait implementation` :t:`type parameter`.
 
 :dp:`fls_fbgwb3pdfgz`
-The :t:`type` of the :t:`assigned operand` of a :t:`division assignment` shall
+The :t:`type` of the :t:`assigned operand` of a :t:`division assignment expression` shall
 implement the :std:`core::ops::DivAssign` :t:`trait` where the :t:`type` of the
 :t:`modifying operand` is the :t:`trait implementation` :t:`type parameter`.
 
 :dp:`fls_8tbxq95x06yt`
-The :t:`type` of the :t:`assigned operand` of a :t:`multiplication assignment`
+The :t:`type` of the :t:`assigned operand` of a :t:`multiplication assignment expression`
 shall implement the :std:`core::ops::MulAssign` :t:`trait` where the :t:`type`
 of the :t:`modifying operand` is the :t:`trait implementation`
 :t:`type parameter`.
 
 :dp:`fls_9oy9zo3x3fy3`
-The :t:`type` of the :t:`assigned operand` of a :t:`remainder assignment` shall
+The :t:`type` of the :t:`assigned operand` of a :t:`remainder assignment expression` shall
 implement the :std:`core::ops::RemAssign` :t:`trait` where the :t:`type` of the
 :t:`modifying operand` is the :t:`trait implementation` :t:`type parameter`.
 
 :dp:`fls_pdgj2xekdead`
-The :t:`type` of the :t:`assigned operand` of a :t:`shift left assignment` shall
+The :t:`type` of the :t:`assigned operand` of a :t:`shift left assignment expression` shall
 implement the :std:`core::ops::ShlAssign` :t:`trait` where the :t:`type` of the
 :t:`modifying operand` is the :t:`trait implementation` :t:`type parameter`.
 
 :dp:`fls_4uoi6k8r7mvc`
-The :t:`type` of the :t:`assigned operand` of a :t:`shift right assignment`
+The :t:`type` of the :t:`assigned operand` of a :t:`shift right assignment expression`
 shall implement the :std:`core::ops::ShrAssign` :t:`trait` where the :t:`type`
 of the :t:`modifying operand` is the :t:`trait implementation`
 :t:`type parameter`.
 
 :dp:`fls_fjaz4m90cagr`
-The :t:`type` of the :t:`assigned operand` of a :t:`subtraction assignment`
+The :t:`type` of the :t:`assigned operand` of a :t:`subtraction assignment expression`
 shall implement the :std:`core::ops::SubAssign` :t:`trait` where the :t:`type`
 of the :t:`modifying operand` is the :t:`trait implementation`
 :t:`type parameter`.
@@ -2667,7 +2667,7 @@ of the :t:`modifying operand` is the :t:`trait implementation`
 .. rubric:: Dynamic Semantics
 
 :dp:`fls_eesn9kuylim`
-The :t:`evaluation` of a :t:`compound assignment` proceeds as follows:
+The :t:`evaluation` of a :t:`compound assignment expression` proceeds as follows:
 
 #. :dp:`fls_4nnqz4etisgw`
    If the :t:`[type]s` of both :t:`[operand]s` are :t:`[integer type]s` or :t:`[floating-point type]s`, then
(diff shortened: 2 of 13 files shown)


src/types-and-traits.rst
```diff
@@ -778,6 +781,9 @@ If the :t:`type` of a :t:`tuple struct field` is a :t:`dynamically sized type`,
 then the :t:`tuple struct field` shall be the last :t:`tuple struct field` in
 the :s:`TupleStructFieldList`.
 
+:dp:`fls_9bp4cKSfW9A0`
```
PLeVasseur: Where: `src/types-and-traits.rst:784-785`.

PR snippet:
> A :dt:`unit struct` is a :t:`struct` without a :s:`RecordStructFieldList` or a :s:`TupleStructFieldList`.

This also feels like the right chapter home. A `unit struct` is a structural property of a struct declaration, so defining it alongside the struct-form rules is more discoverable than leaving it glossary-only.

src/types-and-traits.rst
```diff
@@ -1755,7 +1764,7 @@ ranges, and structures of both :t:`[type]s` are compatible according to the
 rules detailed below.
 
 :dp:`fls_aie0tr62vhw5`
```
PLeVasseur: Where: `src/types-and-traits.rst:1766-1767`, with a nearby use at `src/patterns.rst:89-91`.

PR snippet:
> Two types that :t:`unify` are said to be :t:`unifiable types`.

This seems like a sensible place for the term to appear. The wording now sits next to the type-unification rules that give it meaning, and the use in `src/patterns.rst:89-91` reads naturally against that section.

src/macros.rst
```diff
@@ -25,7 +25,7 @@ distinct forms:
   return a stream of :t:`[lexical element]s`.
 
 :dp:`fls_rnty1c8l5495`
-:t:`[Token]s` are a subset of :t:`[lexical element]s` consumed by :t:`[macro]s`.
+A :dt:`token` is a :t:`lexical element` that can be consumed by a :t:`macro`.
```
PLeVasseur: Where: `src/macros.rst:27-28`.

PR snippet:
> A :dt:`token` is a :t:`lexical element` that can be consumed by a :t:`macro`.

This feels like a good placement. The definition now appears at the start of the macros chapter, right before the text starts relying on token-oriented macro rules, so it is much easier to find than the old glossary-only version.

src/types-and-traits.rst
```diff
@@ -1503,6 +1509,9 @@ share a common storage.
 :dp:`fls_njvdevz0xqc0`
 The :t:`size` of a :t:`recursive type` shall be finite.
 
+:dp:`fls_twC8XhwbI60x`
+A :dt:`thin pointer` is a :t:`value` of  an :t:`indirection type` that refers to a :t:`fixed sized type`.
```
PLeVasseur: Where: `src/types-and-traits.rst:1484-1494,1513`.

PR snippet:
> For a :t:`thin pointer`, the :t:`size` and :t:`alignment` are those of :t:`type`
> :c:`usize`.
>
> A :dt:`thin pointer` is a :t:`value` of  an :t:`indirection type` that refers to a :t:`fixed sized type`.

The new definition is useful, but its placement is slightly awkward because the section already uses `thin pointer` several times before introducing the term. Would we want to move the definition above the size/alignment rules that first rely on it?

src/inline-assembly.rst
```diff
@@ -596,6 +594,9 @@ If a :t:`register argument` has :t:`direction modifier` ``inout`` and an
 :t:`input-output register expression`, then the :t:`input register expression`
 and the :t:`output register expression` shall have the same :t:`type`.
 
+:dp:`fls_FG0EmVgrDUDq`
+:dt:`NaN-boxing` is a technique for encoding :t:`[value]s` using the low order bits of the mantissa of a 64-bit IEEE floating-point ``NaN``.
```
PLeVasseur: Where: `src/inline-assembly.rst:583,598`.

PR snippet:
> On RISC-V architectures, if the :t:`register` belongs to :t:`register class` ``freg``, then :c:`f32` :t:`[value]s` are :t:`NaN-boxed <NaN-boxing>` in a :c:`f64` :t:`value`.
>
> :dt:`NaN-boxing` is a technique for encoding :t:`[value]s` using the low order bits of the mantissa of a 64-bit IEEE floating-point ``NaN``.

This definition looks useful, but it currently appears after the first use of `NaN-boxing` in the preceding bullet. Would we want to move it above that bullet so readers meet the term before it is referenced?

src/glossary.rst
```diff
@@ -4402,12 +4243,6 @@ sync type
 A :dt:`sync type` is a :t:`type` that implements the :std:`core::marker::Sync`
 :t:`trait`.
 
-syntactic category
-^^^^^^^^^^^^^^^^^^
-
-A :dt:`syntactic category` is a nonterminal in the Backus-Naur Form grammar
-definition of the Rust programming language.
-
 tail expression
```
PLeVasseur, line 4243: If these terms are meant to move out of the glossary, they seem to want chapter homes in `src/general.rst`, not deletion. `construct` fits best in `General > Structure` near `src/general.rst:147-160`, where that section repeatedly explains what each chapter says about a construct, while `syntactic category` fits best in `General > Method of Description and Syntax Notation` near `src/general.rst:252-263`, right where PascalCase grammar names are introduced. Could we move them there rather than deleting `syntactic category` outright? Failing that, I think we need to keep `syntactic category` for now, since `construct` at `src/glossary.rst:944-945` still depends on it.

src/glossary.rst
```diff
@@ -4923,12 +4736,6 @@ unnamed constant
 An :dt:`unnamed constant` is a :t:`constant` declared with character 0x5F (low
 line).
 
-unnamed lifetime
-^^^^^^^^^^^^^^^^
-
-An :dt:`unnamed lifetime` is a :t:`lifetime` declared with character 0x5F (low
-line).
-
 unqualified path expression
```
PLeVasseur, line 4736: `elided lifetime` at `src/glossary.rst:1255-1256` still depends on `unnamed lifetime` to distinguish the `'_` case from implicit omission, so removing `unnamed lifetime` leaves that glossary term dangling. If the intent is to migrate this term out of the glossary, its natural chapter home seems to be `src/types-and-traits.rst` in `Lifetime Elision`, where the spec already says a lifetime may be elided implicitly or explicitly and that explicit elision is the ``'_`` lifetime (`src/types-and-traits.rst:3237-3244`; see also the `LifetimeIndication` syntax at `2907-2910`). Could we move `unnamed lifetime` there as a chapter-level `:dt:` definition instead of deleting it outright? Failing that, I think we need to keep the term for now, since `elided lifetime` still depends on it.

src/generics.rst
```diff
@@ -123,8 +123,10 @@ A :t:`generic union` shall use all of its :t:`[type parameter]s` and
 :t:`[lifetime parameter]s` at least once in at least one of its :t:`[field]s`.
 
 :dp:`fls_hyi2jnp38v1n`
-A :t:`generic parameter` is said to constrain an :t:`implementation` if the
-:t:`generic parameter` appears at least once in one of the following:
+A :t:`generic parameter` is said to :dt:`constrain` an :t:`implementation` if it makes the applicability of the :t:`implementation` more narrow.
```
PLeVasseur, line 126: The old wording here, and the Rust Reference, made the bullet list the direct rule for when a generic parameter constrains an implementation. This rewrite replaces that with a broader gloss and moves the actual criterion into a separate sentence about when "a constraint happens", which makes the rule less explicit. Could we keep the direct formulation here, or otherwise say explicitly that a generic parameter constrains the implementation exactly when it appears in one of the following places? If so, `src/glossary.rst:939` may need the same tightening, though that broader glossary phrasing predates this PR.

src/expressions.rst
```diff
@@ -3562,6 +3562,9 @@ The :t:`evaluation` of a :t:`method call expression` proceeds as follows:
 Call Conformance
 ~~~~~~~~~~~~~~~~
 
+:dp:`fls_BBz4Iy7AeaU4`
+:dt:`Call conformance` measures the compatibility between a set of :t:`[argument operand]s` and a set of :t:`[function parameter]s` or :t:`[field]s`.
```
PLeVasseur, line 3566: This new copy fixes the old `a set if` typo, but the glossary copy at `src/glossary.rst:683-685` still says `a set if`, so the two definitions now diverge. Since this PR is aligning glossary and main-text terminology, could we sync the glossary definition too?

src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
    just the language changes that had an impact to the FLS. See the `release
    notes`_ for a full list of changes.
 
+FLS maintenance
+---------------
+
+- Paragraphs added or changed to ensure all glossary terms are also found in the main text:
```
PLeVasseur, line 25: This maintenance summary feels stronger than the current state of the PR. The branch still leaves live references to removed glossary terms at `src/glossary.rst:945`, `1255`, and `3537`, so the claim here seems too strong as written. This block also does not appear to cover the full set of related edits: `fls_rnty1c8l5495` (`src/macros.rst:28`) and `fls_aie0tr62vhw5` (`src/types-and-traits.rst:1767`) are not listed here, and `fls_kv533rntni1x` (`src/patterns.rst:91`) is only mentioned earlier in the changelog. Could we narrow the wording here, or wait to add the stronger summary until those dangling references are resolved?

src/glossary.rst
```diff
@@ -98,12 +93,6 @@ anonymous return type
 An :dt:`anonymous return type` is an :t:`impl trait type` ascribed to a
 :t:`function` return type.
```
PLeVasseur, line 95: If the intent is to migrate this term out of the glossary, the natural destination seems to be `src/types-and-traits.rst` in `Trait Types -> Impl Trait Types`, immediately before the existing `anonymous return type` paragraph. That subsection already says an `impl trait type` may appear in a `function parameter` or `return type` (`src/types-and-traits.rst:1144-1145`), and then names only the return-position case (`src/types-and-traits.rst:1148-1149`). Could we move `anonymous type parameter` there as the parameter-position counterpart, rather than dropping the name entirely? The example at `src/types-and-traits.rst:1178-1179` is still `anonymous_type_parameter`, so the chapter already has the right context.

src/glossary.rst
```diff
@@ -5191,11 +4993,6 @@ whitespace string
 A :dt:`whitespace string` is a string that consists of one or more
 :t:`[whitespace character]s`.
```
PLeVasseur, line 4992: This feels more like a discoverability nit than a broken removal, but `zero-sized type` is still a common Rust layout term and the spec no longer gives it a direct home. Its natural placement seems to be `src/types-and-traits.rst` in `Representation > Type Layout`, right near `src/types-and-traits.rst:1403-1422`, where that section already defines `dynamically sized type`, `alignment`, and `size`. The later unit-tuple and transparent-representation rules (`src/types-and-traits.rst:1477-1478`, `1557-1560`) already rely on the same concept indirectly. Feels like, to me, it'd be good to keep this term and make a home for it in that chapter.

src/types-and-traits.rst
PLeVasseur, line 1519: The new definition is useful, but its placement is slightly awkward because the section already uses `thin pointer` several times before introducing the term. Would we want to move the definition above the size/alignment rules that first rely on it?

src/inline-assembly.rst
PLeVasseur, line 598: This definition looks useful, but it currently comes after the first in-chapter use of `NaN-boxing` in the RISC-V `freg` exception at `src/inline-assembly.rst:582-583`. Would we want to move the `:dt:` paragraph to immediately before the small-register-size rule block that begins at `src/inline-assembly.rst:578`, so readers meet the term before that first use without breaking up the bullet list?

src/glossary.rst
```diff
@@ -1049,11 +1007,6 @@ A :dt:`crate public modifier` is a :t:`visibility modifier` that grants a
 
 See :s:`CratePublicModifier`.
 
-crate root
-^^^^^^^^^^
-
-A :dt:`crate root` is an entry point into a :t:`crate`.
-
 crate root module
```
PLeVasseur, line 1008: This branch removes `crate root` but keeps `crate root module` as the surviving defined/main-text term (`src/glossary.rst:1010-1014`, `src/program-structure-and-compilation.rst:207-215`). I could only find one remaining `crate root` use, the wrapped reference at `src/entities-and-resolution.rst:953-954` (``:t:`crate\nroot```), while nearby and elsewhere the spec already uses `crate root module`. Could we rewrite that survivor to ``:t:`crate root module``` so the terminology is consistent with the retained chapter-defined term?

src/changelog.rst
```diff
@@ -19,6 +19,45 @@ with the change that has been applied due to it.
...
+
+  - :p:`fls_BBz4Iy7AeaU4`
+  - :p:`fls_cxNAtLSLjqbQ`
+  - :p:`fls_hyi2jnp38v1n`
+  - :p:`fls_twC8XhwbI60x`
+  - :p:`fls_9bp4cKSfW9A0`
+  - :p:`fls_Odh6Z7ivD1U4`
+  - :p:`fls_FG0EmVgrDUDq`
+
+- Paragraphs changed to avoid referencing indirect terms:
```
PLeVasseur, line 35: This PR shows a potential process gap: unresolved `:t:` fallout can pass CI (for example `src/glossary.rst:945`, `src/glossary.rst:1255`, `src/glossary.rst:3537`, and wrapped ``:t:`crate\nroot``` at `src/entities-and-resolution.rst:953-954`). Since this tracking effort depends on moving terms into chapter `:dt:` definitions safely, we should have an automated guardrail. Could we track a dedicated lint or CI check that verifies every `:t:` resolves to a live `:dt:` definition, including wrapped forms, rather than relying on manual cleanup?

Details:

- Missing :t: refs do not become Sphinx warnings in this repo.                                                       
- In exts/ferrocene_spec/definitions/__init__.py:136-158, if a definition target is missing, the extension does this:
  - it does not log a warning                                                                                        
  - it does not create a broken link                                                                                 
  - it replaces the role with plain inline text and adds CSS class spec-missing-ref                                  
- For term refs specifically, exts/ferrocene_spec/definitions/terms.py:37-38 returns just nodes.Text(text).          
- The theme then only colors that text as an error in HTML: themes/fls/static/fls.css:761-763.

src/glossary.rst
```diff
@@ -4448,13 +4283,12 @@ A :dt:`textual type` is a :t:`type` class that includes type :c:`char` and type
...
 
-A :dt:`thin pointer` is a :t:`value` of a :t:`thin pointer type`.
+A :dt:`thin pointer` is a :t:`value` of  an :t:`indirection type` that refers to a :t:`fixed sized type`.
 
-thin pointer type
-^^^^^^^^^^^^^^^^^
+token
+^^^^^
 
-A :dt:`thin pointer type` is an :t:`indirection type` that refers to a
```
PLeVasseur, line 4456: This seems workable as-is, but it does leave the pointer terminology a little asymmetric. This PR folds `thin pointer type` into the value-level `thin pointer` (`src/types-and-traits.rst:1513`), while `fat pointer type` remains the surviving type-level term (`src/types-and-traits.rst:1410`; glossary copy at `src/glossary.rst:1458-1461`). If that asymmetry is intentional, no change is needed; otherwise, would we want to either keep both sides type-level or add a lightweight value-level `fat pointer` companion beside `fat pointer type`?

src/glossary.rst
```diff
@@ -4448,13 +4283,12 @@ A :dt:`textual type` is a :t:`type` class that includes type :c:`char` and type
 thin pointer
 ^^^^^^^^^^^^
 
-A :dt:`thin pointer` is a :t:`value` of a :t:`thin pointer type`.
+A :dt:`thin pointer` is a :t:`value` of  an :t:`indirection type` that refers to a :t:`fixed sized type`.
```
PLeVasseur, line 4283: Minor typo: `of  an` has a doubled space here; the chapter copy at `src/types-and-traits.rst:1513` has the same typo. Could we fix both copies?

src/glossary.rst
```diff
@@ -3670,16 +3546,6 @@ A :dt:`record struct pattern` is a :t:`pattern` that matches a
 
 See :s:`RecordStructPattern`.
 
-record struct type
-^^^^^^^^^^^^^^^^^^
-
-A :dt:`record struct type` is the :t:`type` of a :t:`record struct`.
```
PLeVasseur, line 3676: `record struct type` seems to have a natural chapter home in `src/types-and-traits.rst` under `Struct Types`, immediately after the base `struct type` definition at `src/types-and-traits.rst:766-768`. That section already introduces `RecordStructDeclaration` / `RecordStructFieldList` syntax (`src/types-and-traits.rst:738-750`), and the next rules immediately start talking about `record struct field` (`src/types-and-traits.rst:770-777`). If the intent is to migrate glossary terms into chapters, could we move `record struct` and `record struct type` there together instead of deleting `record struct type` outright? Failing that, I think we need to keep `record struct type` for now, since `record struct field` at `src/glossary.rst:3537` still depends on it.

PLeVasseur (changes requested):
Hey @tshepang -- thanks so much for taking the time to do this work!

I think what I found was a mix of real issues and various levels of nits.

Happy to hear your thoughts.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/689/0ff3456e02d31d5ddf9f01b4a0339fc1deadb7b9..51752ff5bc2621be2737b82ce29f972c7c61bbc6)*

src/glossary.rst
tshepang: - nice that you caught this because it was not supposed to be deleted (and I found a few other cases affected in a similar way)
- `construct` in General is just used as an example... made https://github.com/rust-lang/fls/pull/699 to make it more clear

tshepang: da45bbd3662b66b6b587aa3f588888a51cf4b249

src/glossary.rst
tshepang: See da45bbd3662b66b6b587aa3f588888a51cf4b249, which resulted from me noticing (thanks to you) there were duplicate terms... existence of "unnamed lifetime" is already covered by fls_5ZAQ9p7jQuc2

src/generics.rst
tshepang: hm, am thinking this is fine as the new rule is still accurate, and that it is fine for rules to be further tightened with other rules

am happy to change this though if you feel strongly... I feel weakly, but I did this to avoid definitions being defined by lists, instead of being able to stand alone

src/expressions.rst
tshepang: https://github.com/rust-lang/fls/compare/da45bbd3662b66b6b587aa3f588888a51cf4b249..aa8a0676558a4b3f3bfbdd195e62ca2b4afd177b

src/changelog.rst
tshepang: I would not want the PR to be accepted until the maintenace summary claim is reached

the example you gave are mere wording fixes/adjustments, and therefore do not need to be accounted for in the changelog

src/generics.rst
kirtchev-adacore: "Constrain" is a very broad notion, so I do not think it should be made into a term. The former wording acts as an introduction to the various cases enumerated below.

src/inline-assembly.rst
kirtchev-adacore: In my opinion, NaN-boxing is an implementation details, and should not be mentioned in the FLS.

Numerical details are delegated to the corresponding primitive types, and those are currently outside the scope of the FLS.

src/lexical-elements.rst
```diff
@@ -1332,6 +1332,12 @@ A :t:`block comment` is a :t:`comment` that spans one or more :t:`[line]s`.
 :dp:`fls_nayisy85kyq2`
 A :t:`line comment` is a :t:`comment` that spans exactly one :t:`line`.
 
+:dp:`fls_5SVthCNZBcVS`
+A :dt:`syntactic category` is a nonterminal in the Backus-Naur Form grammar definition of the Rust programming language.
+
+:dp:`fls_SonodAQVL5J4`
+A :dt:`construct` is a piece of program text that is an instance of a :t:`syntactic category`.
+
```
kirtchev-adacore, line 1340: These terms have no business being defined in the section for comments. I think they should at the bottom of 1.1.4. Method of Description and Syntax Notation.

What do you think @PLeVasseur ? I was also considering a new section in 2. Lexical Elements, but we are now at a higher level of abstraction. The two rules relate to BNF.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/689/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..aa8a0676558a4b3f3bfbdd195e62ca2b4afd177b)*

src/types-and-traits.rst
```diff
@@ -696,6 +696,9 @@ It is a static error if the :t:`value` of a :t:`discriminant` exceeds the
 maximum :t:`value` of the :t:`type` of the :t:`expression` of a :t:`discriminant
 initializer`.
 
+:dp:`fls_Odh6Z7ivD1U4`
+A :dt:`unit enum variant` is an :t:`enum variant` without a :s:`RecordStructFieldList` or :s:`TupleStructFieldList`.
+
```
kirtchev-adacore, line 701: Do not simply append new rules to the end of a section. You should consider where they go within that section based on the flow of information being introduced.

For instance, this rule should go after 4.5.1:3 An enum variant... since it is closely related to enum variants, and introduces a new form of an enum variant. Placing it after a rule about exceeding the maximum value of a type is bizarre, as it has no relation to the immediately preceding rules.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/689/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..aa8a0676558a4b3f3bfbdd195e62ca2b4afd177b)*

src/types-and-traits.rst
```diff
@@ -778,6 +787,9 @@ If the :t:`type` of a :t:`tuple struct field` is a :t:`dynamically sized type`,
 then the :t:`tuple struct field` shall be the last :t:`tuple struct field` in
 the :s:`TupleStructFieldList`.
 
+:dp:`fls_9bp4cKSfW9A0`
+A :dt:`unit struct` is a :t:`struct` without a :s:`RecordStructFieldList` or a :s:`TupleStructFieldList`.
+
```
kirtchev-adacore, line 792: This rule should come after :dp:`fls_mdejdc0nNgyO`, since it is introducing yet another kind of struct.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/689/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..aa8a0676558a4b3f3bfbdd195e62ca2b4afd177b)*

src/types-and-traits.rst
kirtchev-adacore: Section 4.7. Indirection Types should be given its own Legality Rules, and this rule should be relocated there.

On a side note, is "thin pointer" used anywhere else in the FLS?

kirtchev-adacore (changes requested):

src/generics.rst
tshepang: @kirtchev-adacore am in favor of that... probably kept since it was already made a term before this pr

src/glossary.rst
tshepang: I think we should not have terms that are not referenced by any rule. If it happens that some rule in future needs the term to exist, we can bring it back.

Otherwise, where do you think we should use it, so we can keep the term? Can we do that in a separate PR?

src/glossary.rst
tshepang: also 224318e193c

src/lexical-elements.rst
tshepang: I don't know how it happened, placing those there... fixed in 224318e193c

src/types-and-traits.rst
tshepang: bbe72da5b22

src/types-and-traits.rst
tshepang: 13e45c7a7675b2e6790ed440f7b2dc8945d07e27

src/glossary.rst
tshepang: 2f227a2525c989ed2bfc7a39a57f95c1bcd4cd7e

src/types-and-traits.rst
tshepang: handled by https://github.com/rust-lang/fls/pull/715

src/inline-assembly.rst
tshepang: we decided on removing risc-v from FLS (until the platform is qualified), so this will go away

src/glossary.rst
tshepang: 63ee6cc0594d0a8310b1ac43c7540b09fffd887b

src/glossary.rst
tshepang: handled in https://github.com/rust-lang/fls/pull/715... the term is no longer removed here

src/glossary.rst
tshepang: no longer applies since rebase on https://github.com/rust-lang/fls/pull/715

src/glossary.rst
tshepang: this removal does not exist in the pr (and am failing see how it did exist)

src/changelog.rst
tshepang: 3b3bd6047de775bd252bfffbc248a1de85cd0287

src/generics.rst
kirtchev-adacore: I think this paragraph should be expressed in terms of a "generic constraint". If you search the FLS, you will find two other very different instances of "to contain" / "constraint".

How about:

A generic constraint is a limitation imposed on an implementation by a generic parameter.

A generic parameter constrains (link to "generic constraint") an implementation if ...

It is a static error if a lifetime parameter ... without constraining (link to "generic constraint") the implementation.

src/inline-assembly.rst
```diff
@@ -580,9 +580,7 @@ If a :t:`value` has a smaller size than the :t:`register` it is allocated in,
 then
 
 * :dp:`fls_drg7v8hxb5ca`
-  On RISC-V architectures, if the :t:`register` belongs to :t:`register class`
-  ``freg``, then :c:`f32` :t:`[value]s` are :t:`NaN-boxed <NaN-boxing>`. in a
-  :c:`f64` :t:`value`.
+  On RISC-V architectures, if the :t:`register` belongs to :t:`register class` ``freg``, then :c:`f32` :t:`[value]s` are :t:`NaN-boxed <NaN-boxing>` in a :c:`f64` :t:`value`.
```
kirtchev-adacore, line 583: We decided to remove the mentions of RISCV, so this paragraph should be removed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/689/a70c80edeeb38e20dea1040a5773705e3eeeab85..d6707f82286f945f4c769f47654eab0639063afa)*

src/inline-assembly.rst
kirtchev-adacore: Reminding you to remove this paragraph.

src/types-and-traits.rst
```diff
@@ -1113,8 +1125,10 @@ An :t:`impl trait type` is a :t:`type` that implements a :t:`trait`, where the
 :t:`type` is known at compile time.
 
 :dp:`fls_ieyqx5vzas2m`
-An :t:`impl trait type` shall appear only within a :t:`function parameter` or
-the :t:`return type` of a :t:`function`.
+An :t:`impl trait type` is either an :t:`anonymous type parameter` or an :t:`anonymous return type`.
```
kirtchev-adacore, line 1128: What happened here? There are now two definition of "impl trait type", this one, and the one from fls_a6zlvyxpgsew above.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/689/a70c80edeeb38e20dea1040a5773705e3eeeab85..d6707f82286f945f4c769f47654eab0639063afa)*

PR #693, opened 2026-04-14
document if let guards

tshepang:


src/expressions.rst
```diff
@@ -4645,7 +4645,22 @@ Match Expressions
        OuterAttributeOrDoc* Pattern MatchArmGuard?
 
    MatchArmGuard ::=
-       $$if$$ Operand
+       $$if$$ (Operand | MatchGuardChain)
+
+   MatchGuardChain ::=
+       MatchGuardCondition ($$&&$$ MatchGuardCondition)*
```
jyn514: this has `&&` but not `||`, is that intentional? I know that's true for `if let` guards but I'm not sure if it's also true for match guards.

jyn514 (commented):
c.f. https://doc.rust-lang.org/reference/expressions/match-expr.html#match-guards

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

src/expressions.rst
tshepang: it is intentional... `||` is not allowed in [both](https://github.com/rust-lang/reference/blob/a65a85110a50d312de798be5da6c5d1624a49a40/src/expressions/match-expr.md?plain=1#L235) [cases](https://doc.rust-lang.org/reference/expressions/if-expr.html#r-expr.if.chains.or)

jyn514 (approved):

src/expressions.rst
```diff
@@ -4645,7 +4645,22 @@ Match Expressions
...
+       (MatchGuardExpression | OuterAttributeOrDoc* MatchGuardLetPattern)
+
+   MatchGuardLetPattern ::=
+       $$let$$ Pattern $$=$$ MatchGuardExpression
+
+   MatchGuardExpression ::=
+       Expression
+
+:dp:`fls_UlxLrpyPlVmv`
+A :dt:`match guard expression` is any expression in category :s:`Expression`, except :s:`LazyBooleanExpression`, :s:`RangeFromToExpression`, :s:`RangeFromExpression`, :s:`RangeInclusiveExpression`, :s:`AssignmentExpression`, and :s:`CompoundAssignmentExpression`.
```
kirtchev-adacore:
```suggestion
A :dt:`match guard expression` is any :t:`expression` in category :s:`Expression`, except
- :s:`AssignmentExpression`
- :s:`CompoundAssignmentExpression`
- :s:`LazyBooleanExpression`
- :s:`RangeFromExpression`
- :s:`RangeFromToExpression`
- :s:`RangeInclusiveExpression`
```
It is cleaner to use lists. The list items should be alphabetized.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

src/expressions.rst
```diff
@@ -4715,6 +4730,15 @@ match the :t:`[subject expression]'s` :t:`type`.
 The :t:`value` of a :t:`match expression` is the :t:`value` of the :t:`operand`
 of the selected :t:`match arm`.
 
+:dp:`fls_DT4N2rr6wpvZ`
```
kirtchev-adacore: This paragraph should follow "6.18:8 A match arm guard ...".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

src/expressions.rst
```diff
@@ -4645,7 +4645,22 @@ Match Expressions
        OuterAttributeOrDoc* Pattern MatchArmGuard?
 
    MatchArmGuard ::=
-       $$if$$ Operand
+       $$if$$ (Operand | MatchGuardChain)
```
kirtchev-adacore: I wonder whether these new non-terminal should have "MatchArm" in their names, as opposed to just "Match".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

src/expressions.rst
```diff
@@ -4715,6 +4730,15 @@ match the :t:`[subject expression]'s` :t:`type`.
 The :t:`value` of a :t:`match expression` is the :t:`value` of the :t:`operand`
 of the selected :t:`match arm`.
 
+:dp:`fls_DT4N2rr6wpvZ`
+A :dt:`match guard chain` is a set of conditions that should each evaluate to ``true`` in the case of :t:`[match guard expression]s` or have a positive match in the case of a :t:`[match guard let pattern]s` for the :t:`match arm` to be selected.
```
kirtchev-adacore: From the way this rule is worded, it appears that a single positive match is sufficient to evaluate the guard to true. How about
```suggestion
A :dt:`match guard chain` is a set of conditions that must each evaluate to ``true`` in the case of :t:`[match guard expression]s`, or must each produce a positive match in the case of a :t:`[match guard let pattern]s` for the :t:`match arm` to be selected.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

src/expressions.rst
```diff
@@ -4715,6 +4730,15 @@ match the :t:`[subject expression]'s` :t:`type`.
...
+
+:dp:`fls_AAuyKfxLgJ43`
+A :dt:`match guard let pattern` is a construct that evaluates if a :t:`match guard expression` matches the specified :t:`pattern`.
```
kirtchev-adacore:
```suggestion
A :dt:`match guard let pattern` is evaluated when its :t:`match guard expression` matches the specified :t:`pattern`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

src/expressions.rst
```diff
@@ -4715,6 +4730,15 @@ match the :t:`[subject expression]'s` :t:`type`.
...
+
+:dp:`fls_uCDQMkWx5OMS`
+Each :t:`let binding` in a :t:`match guard let pattern` is in scope for the rest of the :t:`match arm guard` as well as the :t:`match arm body`.
```
kirtchev-adacore:
```suggestion
Each :t:`let binding` introduced in a :t:`match guard let pattern` is :t:`in scope` for the rest of the :t:`match arm guard` as well as the :t:`match arm body`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

src/expressions.rst
```diff
@@ -4767,11 +4791,6 @@ The :t:`evaluation` of a :t:`match arm matcher` proceeds as follows:
 #. :dp:`fls_yk8l9zjh7i0d`
    Otherwise the :t:`match arm matcher` fails.
 
-:dp:`fls_sbtx1l6n2tp2`
-The :t:`evaluation` of a :t:`match arm guard` evaluates its :t:`operand`. A
```
kirtchev-adacore, line 4771: Hmm, this paragraph is actually useful. In fact, there Dynamic Semantics section should be updated to account for match guard let patterns.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/f5c48f51906cc12b5745ced6658d96ea0850552a..79c291048f0ffdbcc74b745439d468be99436e97)*

kirtchev-adacore (changes requested):

src/expressions.rst
tshepang: thought it was redundant since `fls_wkh5wztauwhu` says the same thing

src/expressions.rst
```diff
@@ -4645,7 +4645,35 @@ Match Expressions
        OuterAttributeOrDoc* Pattern MatchArmGuard?
 
    MatchArmGuard ::=
-       $$if$$ Operand
+       $$if$$ (Operand | MatchArmGuardChain)
+
+   MatchArmGuardChain ::=
+       MatchArmGuardCondition ($$&&$$ MatchArmGuardCondition)*
+
+   MatchArmGuardCondition ::=
```
PLeVasseur, line 4653: The new grammar now admits two different kinds of guard condition, but the old single-`bool` rules are still in place.

For `if let PAT = EXPR`, `EXPR` is a scrutinee matched against `PAT`; `EXPR` itself is not required to have type `bool`. Only expression-valued guard conditions are `bool`-typed. As written, that seems to conflict with both the existing legality rule in `src/expressions.rst` (`fls_bzhz5wjd90ii`) and the stale inference rule in `src/types-and-traits.rst` (`fls_st9onPgDrc8y`), which still infer all match-arm-guard operands with expected type `bool`.

Could this be split the same way the Reference and the existing FLS already distinguish ordinary boolean conditions from `if let` / `while let` scrutinees? One possible shape would be a distinct guard-scrutinee rule with the pattern's type as its expected type.

Support:
- Reference grammar `MatchGuardCondition`
- Reference grammar `MatchGuardScrutinee`
- Reference `expr.match.guard.condition`
- FLS `src/expressions.rst` `fls_bzhz5wjd90ii`
- FLS `src/types-and-traits.rst` `fls_st9onPgDrc8y`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/type-inference.rs`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/typeck.rs`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..eba53099764a6a3411e3aadf1cbb9f8496c81300)*

src/expressions.rst
```diff
@@ -4715,6 +4746,12 @@ match the :t:`[subject expression]'s` :t:`type`.
 The :t:`value` of a :t:`match expression` is the :t:`value` of the :t:`operand`
 of the selected :t:`match arm`.
 
+:dp:`fls_AAuyKfxLgJ43`
+A :dt:`match arm guard let pattern` is evaluated when its :t:`match arm guard expression` matches the specified :t:`pattern`.
+
+:dp:`fls_uCDQMkWx5OMS`
```
PLeVasseur, line 4752: I think this stabilization is still incomplete outside `src/expressions.rst`.

The new local rule says bindings introduced by a guard `let` are in scope for the rest of the guard and the arm body, but the shared scope / destructor chapters still describe the older model: `src/entities-and-resolution.rst` `fls_xbnki64un70v` only covers arm-pattern bindings, and `src/ownership-and-deconstruction.rst` `fls_lbsfhg42yiqy`, `fls_fnvr5w2wzxns`, and `fls_ptk6yibqyfzi` still talk about a single guard operand rather than pattern-matching guard conditions with arm-scope bindings and temporaries.

Upstream needed matching updates in `names/scopes` and `destructors` for the same feature (`rust-lang/reference#1957`). Could this PR update those shared chapters too? Otherwise the new local rule here will not match the shared scope / drop-scope text.

Support:
- Reference PR `rust-lang/reference#1957`
- Reference `names.scopes.pattern-bindings.match-guard-let`
- Reference `expr.match.guard.chains.bindings`
- Reference `destructors.scope.bindings.match-arm`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/scope.rs`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/scoping-consistency.rs`
- rustc test `tests/ui/drop/if-let-guards.rs`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..eba53099764a6a3411e3aadf1cbb9f8496c81300)*

src/expressions.rst
```diff
@@ -4715,6 +4746,12 @@ match the :t:`[subject expression]'s` :t:`type`.
 The :t:`value` of a :t:`match expression` is the :t:`value` of the :t:`operand`
 of the selected :t:`match arm`.
 
+:dp:`fls_AAuyKfxLgJ43`
```
PLeVasseur, line 4749: I think FLS still needs an explicit guard-specific rule for when arm-pattern bindings become values.

Today the shared pattern rules say by-value bindings move or copy during pattern matching (`src/patterns.rst` `fls_pxvtqxke1enp`), but the stabilized guard behavior is different for arm-pattern bindings: during guard evaluation they are observed through shared references, and only after the guard succeeds are values moved or copied from the scrutinee into the arm bindings. Without that exception, the spec still reads as though a by-value arm binding may move before guard evaluation.

Could we add the guard-specific rule from the Reference here (or in the shared pattern / guard text), and tie it back to the existing immutability rule for match-arm-guard bindings? That would also explain why mutation through those bindings is rejected while the guard runs.

Support:
- Reference `expr.match.guard.shared-ref`
- Reference `expr.match.guard.value`
- Reference `expr.match.guard.no-mutation`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/guard-mutability-2.rs`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/move-guard-if-let.rs`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/move-guard-if-let-chain.rs`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..eba53099764a6a3411e3aadf1cbb9f8496c81300)*

src/expressions.rst
PLeVasseur: I agree the Dynamic Semantics section needs an explicit update here, not just a revised success criterion.

As written, `fls_DT4N2rr6wpvZ` still reads more like an unordered condition for when a guard succeeds. The Reference specifies this operationally: once the arm pattern matches, guard conditions are evaluated left-to-right; evaluation stops at the first `false` or failed `let` match; later conditions are not evaluated; and if the guard fails, matching resumes, including later `|` alternatives in the same arm.

Could we make those parts explicit in Dynamic Semantics, perhaps by replacing the removed `fls_sbtx1l6n2tp2` with a chain-aware evaluation rule?

Support:
- Reference `expr.match.guard.behavior`
- Reference `expr.match.guard.next`
- Reference `expr.match.guard.chains.order`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/move-guard-if-let-chain.rs`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/drop-order.rs`

PLeVasseur (changes requested):

src/expressions.rst
```diff
@@ -4645,7 +4645,35 @@ Match Expressions
...
+       MatchArmGuardCondition ($$&&$$ MatchArmGuardCondition)*
+
+   MatchArmGuardCondition ::=
+       (MatchArmGuardExpression | OuterAttributeOrDoc* MatchArmGuardLetPattern)
+
+   MatchArmGuardLetPattern ::=
+       $$let$$ Pattern $$=$$ MatchArmGuardExpression
+
+   MatchArmGuardExpression ::=
+       Expression
```
kirtchev-adacore, line 4660:
```suggestion
   MatchArmGuardExpression ::=
       Expression

   MatchArmGuardLetPattern ::=
       $$let$$ Pattern $$=$$ MatchArmGuardExpression
```
I reordered these so that the order reflects the order in `MatchArmGuardCondition`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..eba53099764a6a3411e3aadf1cbb9f8496c81300)*

src/expressions.rst
kirtchev-adacore: I think `MatchArmGuardCondition` needs to be changed as follows:
```
   MatchArmGuardCondition ::=
     (Operand | OuterAttributeOrDoc* MatchArmGuardLetPattern)
```
This way a leading expression which starts a guard chain still needs to evaluate to `bool` as per `fls_bzhz5wjd90ii`.

I think `fls_bzhz5wjd90ii` itsled should also be changed to

The :t:`type` of an :t:`operand` of a :t:`match arm guard` shall be :t:`type`
:c:`bool`.

(I used "an" instead of "the" since a guard chain can now introduce multiple operands)

**This suggestion is no longer relevant.**

kirtchev-adacore: Actually `MatchArmGuard` needs to be changed too, and one of the productions thrown out:
```
MatchArmGuard ::=
    $$if$$ MatchArmGuardChain

MatchArmGuardChain ::=
    MatchArmGuardCondition ($$&&$$ MatchArmGuardCondition)*

MatchArmGuardCondition ::=
    Operand
  | OuterAttributeOrDoc* MatchArmGuardLetPattern
```
This way "pure" expressions (operands) must evaluate to `bool`, and the grammar now handles a starting operand or a starting let pattern.

**This suggestion is no longer relevant.**

src/expressions.rst
```diff
@@ -4715,6 +4746,12 @@ match the :t:`[subject expression]'s` :t:`type`.
 The :t:`value` of a :t:`match expression` is the :t:`value` of the :t:`operand`
 of the selected :t:`match arm`.
 
+:dp:`fls_AAuyKfxLgJ43`
+A :dt:`match arm guard let pattern` is evaluated when its :t:`match arm guard expression` matches the specified :t:`pattern`.
```
kirtchev-adacore, line 4750: "Evaluation" is a dynamic semantic, so it has no business in the Legality Rules section.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..eba53099764a6a3411e3aadf1cbb9f8496c81300)*

src/expressions.rst
kirtchev-adacore: `fls_sbtx1l6n2tp2` needs to be changed into:

The :t:`evaluation` of a :t:`match arm guard` proceeds as follows:
1. Each :t:`match arm guard condition` is evaluated in left-to-right order as follows:
    1. If the :t:`match arm guard condition` is an :t:`operand`, then
        1. The :t:`operand` is evaluated.
        2. If the :t:`operand` evaluated to ``false``, then the :t:`match arm guard` evaluates to ``false``.
    2. If the :t:`match arm guard condition` is a :t:`match arm guard let pattern`, then
        1. The :t:`pattern` of the :t:`match arm guard let pattern` is evaluated.
        2. If the :t:`pattern` failed to match, then the :t:`match arm guard` evaluates to ``false``.
2. The :t:`match arm guard` evaluates to ``true``.

**This suggestion is no longer relevant.**

kirtchev-adacore (changes requested):

src/expressions.rst
PLeVasseur: I think this needs a source-level split between ordinary match-arm guard operands, expression-valued guard-chain condition operands, and guard-let scrutinees.

The reason is semantic. The Reference distinguishes these roles:

- `MatchConditions -> MatchGuardChain | Expression`
- `MatchGuardCondition -> Expression except ExcludedMatchConditions | let Pattern = MatchGuardScrutinee`
- `MatchGuardScrutinee -> Expression except ExcludedMatchConditions`

Only expression-valued guard operands/conditions are required to have type `bool`. A guard-let scrutinee is matched against the related pattern and is not itself required to have type `bool`.

As written, `MatchArmGuardExpression` appears to cover ordinary guard expressions, chain condition expressions, and guard-let scrutinees. That leaves the existing FLS rules stale or ambiguous: `src/expressions.rst` `fls_bzhz5wjd90ii` still speaks about the `operand` of a match arm guard being `bool`, and `src/types-and-traits.rst` `fls_st9onPgDrc8y` still infers all match-arm-guard operands with expected type `bool`.

One concrete FLS shape would be this disjoint split. I do not care about the exact category names, but I think the roles need to remain distinct. This shape intentionally omits the Reference's guard-let attribute position for Rust 2021 FLS because stable `rustc +1.95.0 --edition=2021` rejects attributes before guard-let patterns with `E0658`:

```rst
   MatchArmGuard ::=
       $$if$$ MatchArmGuardConditions

   MatchArmGuardConditions ::=
       MatchArmGuardChain
     | MatchArmGuardOperand

   MatchArmGuardOperand ::=
       Operand

   MatchArmGuardChain ::=
       MatchArmGuardConditionOperandBeforeLet* MatchArmGuardLetPattern MatchArmGuardConditionAfterLet*

   MatchArmGuardConditionOperandBeforeLet ::=
       MatchArmGuardConditionOperand $$&&$$

   MatchArmGuardConditionAfterLet ::=
       $$&&$$ MatchArmGuardCondition

   MatchArmGuardCondition ::=
       MatchArmGuardConditionOperand
     | MatchArmGuardLetPattern

   MatchArmGuardConditionOperand ::=
       Expression

   MatchArmGuardLetPattern ::=
       $$let$$ Pattern $$=$$ MatchArmGuardScrutinee

   MatchArmGuardScrutinee ::=
       Expression
```

With this shape, expression-only guards such as `if a && b` remain ordinary `MatchArmGuardOperand`s, while chain-specific rules apply only when a guard-let pattern is present. A more Reference-shaped grammar can also work, provided the final text still separates ordinary guard operands, chain condition operands, and guard-let scrutinees for typing, inference, scope, drop-scope, temporary-scope, and dynamic-semantics purposes. For FLS Rust 2021, I do not think the grammar should require the Reference's guard-let attribute position unless stable rustc accepts it or the divergence is explicitly documented.

A bare `Operand` for every chain condition is too broad in the FLS because `Operand ::= Expression` and would admit categories the Reference excludes inside guard chains.

Target: `src/expressions.rst`, immediately after the syntax block. Replace the current `fls_UlxLrpyPlVmv` paragraph and its six list-item paragraphs (`fls_XADcpJBUxSfv`, `fls_gfHe2Cy6WXsK`, `fls_QQep7FKA1EQX`, `fls_Wepy5R7FZQPU`, `fls_imEIc7PUUO1x`, `fls_fs4ZpXjt0Wqt`) with syntax-subset definitions. The exclusions belong on chain operands and guard-let scrutinees, not on ordinary guards. Since the syntax categories are defined by the surrounding `.. syntax::` block, these subset paragraphs should use `:s:` links rather than introducing duplicate syntax definitions with `:ds:`:

```rst
:dp:`<new-id>`
A :s:`MatchArmGuardConditionOperand` is any :t:`expression` in category
:s:`Expression`, except expressions in any of the following categories:

* :dp:`<new-id>`
  :s:`AssignmentExpression`
* :dp:`<new-id>`
  :s:`CompoundAssignmentExpression`
* :dp:`<new-id>`
  :s:`LazyBooleanExpression`
* :dp:`<new-id>`
  :s:`RangeFromExpression`
* :dp:`<new-id>`
  :s:`RangeFromToExpression`
* :dp:`<new-id>`
  :s:`RangeInclusiveExpression`

:dp:`<new-id>`
A :s:`MatchArmGuardScrutinee` is any :t:`expression` in category
:s:`Expression`, except expressions in any of the following categories:

* :dp:`<new-id>`
  :s:`AssignmentExpression`
* :dp:`<new-id>`
  :s:`CompoundAssignmentExpression`
* :dp:`<new-id>`
  :s:`LazyBooleanExpression`
* :dp:`<new-id>`
  :s:`RangeFromExpression`
* :dp:`<new-id>`
  :s:`RangeFromToExpression`
* :dp:`<new-id>`
  :s:`RangeInclusiveExpression`
```

The three excluded range categories above intentionally mirror the Reference's excluded range forms for match-guard chains: FLS `RangeFromToExpression` maps to the Reference `RangeExpr`, while `RangeFromExpression` and `RangeInclusiveExpression` map directly. I do not think `RangeFullExpression`, `RangeToExpression`, or `RangeToInclusiveExpression` should be added unless the Reference's `ExcludedMatchConditions` changes.

Target: `src/expressions.rst`, Legality Rules. With a split along these lines, insert the new bucket paragraphs immediately after `fls_hs1rr54hu18w`, then replace `fls_bzhz5wjd90ii` with rules for the two bool-typed cases and the guard-let pattern/scrutinee relationship. Because the corresponding terms should be defined in `src/glossary.rst`, these main-text paragraphs use `:t:` rather than `:dt:`:

```rst
:dp:`fls_DT4N2rr6wpvZ`
A :t:`match arm guard chain` is a sequence of one or more
:t:`[match arm guard condition]s`, separated by ``&&``, that contains at least
one :t:`match arm guard let pattern`.

:dp:`fls_Uv7QsP2nKcDx`
A :t:`match arm guard operand` is the :t:`expression` in category
:s:`MatchArmGuardOperand` of a :t:`match arm guard` that does not have a
:t:`match arm guard chain`.

:dp:`fls_Er5TmL9bXwVq`
A :t:`match arm guard condition operand` is the :t:`expression` in category
:s:`MatchArmGuardConditionOperand` of a :t:`match arm guard condition`.

:dp:`fls_Yq4NhC8sRpJm`
A :t:`match arm guard scrutinee` is the :t:`expression` in category
:s:`MatchArmGuardScrutinee` of a :t:`match arm guard let pattern`.

:dp:`fls_bzhz5wjd90ii`
The :t:`type` of a :t:`match arm guard operand` shall be :t:`type` :c:`bool`.

:dp:`fls_kM6dZv3PaNqT`
The :t:`type` of a :t:`match arm guard condition operand` shall be :t:`type`
:c:`bool`.

:dp:`fls_Lc8EwS5rYpBn`
The :t:`expected type` of the :t:`pattern` of a
:t:`match arm guard let pattern` is the :t:`type` of its
:t:`match arm guard scrutinee`.
```

Target: `src/types-and-traits.rst`, `fls_st9onPgDrc8y`. Replace the old inference step with context-specific inference:

```rst
#. :dp:`fls_st9onPgDrc8y`
   Infer each :t:`match arm guard operand` and each
   :t:`match arm guard condition operand` with :t:`expected type` :c:`bool`.

#. :dp:`fls_vQ3tHg9XmDkF`
   Infer each :t:`match arm guard scrutinee` with the :t:`expected type` set to
   the :t:`type` of the :t:`pattern` of the related
   :t:`match arm guard let pattern`.
```

Target: `src/glossary.rst`, near the existing `match arm guard` entry. Add entries for `match arm guard chain`, `match arm guard condition`, `match arm guard condition operand`, `match arm guard let pattern`, `match arm guard operand`, and `match arm guard scrutinee`, preserving glossary alphabetical order.

Target: `src/changelog.rst`, Rust 1.95.0 entry for `Stabilize if let guards on match arms`. We can use the final source shape to replace the current accounting with one consolidated changelog block. We should not list `MatchArmGuardExpression` or `fls_UlxLrpyPlVmv`/its list-item IDs if the final update uses separate categories for these roles. We should not list `fls_uCDQMkWx5OMS`; that PR-local scope paragraph should be deleted before merge, and the shared scope/drop updates are accounted for by the `src/entities-and-resolution.rst` and `src/ownership-and-deconstruction.rst` IDs in the consolidated block. We should list the changed pattern paragraphs `fls_72JHo343O7jp` and `fls_6bwTtGKb7ba7`, and list `let binding` as an updated glossary entry. The new glossary entries are on an informational page, but the changelog still records them as `New glossary entries` with `:t:` references.

src/expressions.rst
PLeVasseur: I think this should be resolved by the main grammar split rather than by reordering alone.

The issue is not just production order. `MatchArmGuardExpression` is currently doing too much: it covers ordinary guard expressions, expression-valued chain conditions, and guard-let scrutinees, even though those roles need different typing and inference rules.

src/expressions.rst
PLeVasseur: This sentence is describing evaluation, so I think it belongs in Dynamic Semantics rather than Legality Rules.

It also calls the right-hand side a `match arm guard expression`, which risks conflating a guard-let scrutinee with a bool-typed guard condition. The structural legality rule should identify the pieces; the dynamic rule should say that the scrutinee is evaluated and then pattern matching is performed.

Target: `src/expressions.rst`, Legality Rules, `fls_AAuyKfxLgJ43`. Replace the legality sentence with a structural definition. Because this term is defined in `src/glossary.rst`, this main-text paragraph uses `:t:` rather than `:dt:`:

```rst
:dp:`fls_AAuyKfxLgJ43`
A :t:`match arm guard let pattern` is a :t:`match arm guard condition` that
consists of a :t:`pattern` and a :t:`match arm guard scrutinee`.
```

The evaluation steps belong in Dynamic Semantics, not in this legality-rule paragraph.

src/expressions.rst
PLeVasseur: I think the Dynamic Semantics section needs a chain-aware guard evaluation rule, a bit more than what @kirtchev-adacore might've been suggesting.

The Reference specifies this operationally:

- after an arm pattern matches, the guard is evaluated;
- guard conditions are evaluated left-to-right;
- evaluation stops at the first `false` condition operand or failed guard-let match;
- later guard conditions are not evaluated;
- successful guard-let matches introduce bindings for later conditions and the arm body;
- if the guard fails, matching continues, including later `|` alternatives in the same arm.

Current FLS `fls_4dv7x9nh2h4e` evaluates the arm pattern once and then evaluates the guard once. That does not seem to capture the Reference behavior for self-overlapping or-patterns where guard failure tests later `|` alternatives in the same arm. Also, the PR removes `fls_sbtx1l6n2tp2` instead of replacing it with a chain-aware rule.

Target: `src/expressions.rst`, Dynamic Semantics. Replace the deleted `fls_sbtx1l6n2tp2` with an operand-and-chain-aware guard rule. The exact category names can follow the final grammar shape, but the dynamic rule needs to cover ordinary guard operand evaluation, guard-let chain evaluation in source order, short-circuiting after `false` or a failed `let` match, binding introduction after successful guard-let pattern matching, and retry after guard failure including later `|` alternatives in the same arm.

The example below assumes a grammar that separates ordinary guard operands from guard-let chains. If the final grammar uses different category names, we can still adjust the terms while preserving the same evaluation rules.

```rst
:dp:`fls_sbtx1l6n2tp2`
The :t:`evaluation` of a :t:`match arm guard` proceeds as follows:

#. :dp:`fls_Dp8aJr2LqYwS`
   If the :t:`match arm guard` has a :t:`match arm guard operand`, then:

   #. :dp:`fls_Wm4QvN7xTcRb`
      The :t:`match arm guard operand` is evaluated.

   #. :dp:`fls_Ks9FdL3pVxQa`
      If the :t:`match arm guard operand` evaluates to ``false``, then the
      :t:`match arm guard` evaluates to ``false`` and its :t:`evaluation` is
      complete.

   #. :dp:`fls_Rt5XcG8nMbPw`
      If the :t:`match arm guard operand` evaluates to ``true``, then the
      :t:`match arm guard` evaluates to ``true`` and its :t:`evaluation` is
      complete.

#. :dp:`fls_Yh2NqT6vLsCk`
   If the :t:`match arm guard` has a :t:`match arm guard chain`, then:

   #. :dp:`fls_Cw2RsH7pVnQx`
      Each :t:`match arm guard condition` of the chain is evaluated in source
      order as follows:

      #. :dp:`fls_Pb7MzD4wQkXe`
         If the :t:`match arm guard condition` is a
         :t:`match arm guard condition operand`, then:

         #. :dp:`fls_Fj8QpM3dLsVt`
            The :t:`match arm guard condition operand` is evaluated.

         #. :dp:`fls_Qf4VnH8sMxRa`
            If the :t:`match arm guard condition operand` evaluates to
            ``false``, then the :t:`match arm guard` evaluates to ``false`` and
            no later :t:`[match arm guard condition]s` are evaluated.

         #. :dp:`fls_Kz7NvR2xYpWm`
            If the :t:`match arm guard condition operand` evaluates to
            ``true``, then evaluation continues with the next
            :t:`match arm guard condition`.

      #. :dp:`fls_Zc6TbK2wLpNd`
         If the :t:`match arm guard condition` is a
         :t:`match arm guard let pattern`, then:

         #. :dp:`fls_Bd9TmL4qVcNs`
            The :t:`match arm guard scrutinee` is evaluated.

         #. :dp:`fls_Rp5HxK8wJqYd`
            :t:`Pattern matching` is performed with the :t:`pattern` of the
            :t:`match arm guard let pattern` against the resulting :t:`value`.

         #. :dp:`fls_Vr9MqJ5xNsYd`
            If the :t:`pattern matching` fails, then the
            :t:`match arm guard` evaluates to ``false`` and no later
            :t:`[match arm guard condition]s` are evaluated.

         #. :dp:`fls_Nt6QmC3sLpXz`
            If the :t:`pattern matching` succeeds, then any :t:`[binding]s`
            introduced by the :t:`pattern` are bound according to the
            :t:`binding pattern` semantics, and evaluation continues with the
            next :t:`match arm guard condition`.

   #. :dp:`fls_Av4JyT9nMkRp`
      If every :t:`match arm guard condition` of the chain has been evaluated
      without the :t:`match arm guard` evaluating to ``false``, then the
      :t:`match arm guard` evaluates to ``true``.
```

Target: `src/expressions.rst`, the matcher rule at `fls_4dv7x9nh2h4e`. The matcher also needs to preserve retry across later `|` alternatives in the same arm. Source shape:

```rst
:dp:`fls_Nc6JvS9rFaLm`
When the :t:`pattern` of a :t:`match arm matcher` is an :t:`or-pattern`, each
:t:`pattern-without-alternation` of the :t:`or-pattern` is considered in source
order during :t:`evaluation` of the :t:`match arm matcher`. Otherwise, the
:t:`pattern` itself is considered.

:dp:`fls_4dv7x9nh2h4e`
The :t:`evaluation` of a :t:`match arm matcher` proceeds as follows:

#. :dp:`fls_Qr3TcH8vNpKs`
   Each :t:`pattern` determined by :p:`fls_Nc6JvS9rFaLm` is considered in
   source order as follows:

   #. :dp:`fls_Lm5WqZ2dJvYp`
      If :t:`pattern matching` with the :t:`pattern` and the :t:`value` of the
      :t:`subject expression` fails, then evaluation continues with the next
      :t:`pattern`.

   #. :dp:`fls_Bx8KsV4nQdRt`
      If :t:`pattern matching` with the :t:`pattern` succeeds, then:

      #. :dp:`fls_Mv2YpG7sLcHw`
         If the :t:`match arm matcher` has no :t:`match arm guard`, then the
         :t:`match arm matcher` succeeds and its :t:`evaluation` is complete.

      #. :dp:`fls_Xd4LqN9vTrPm`
         If the :t:`match arm matcher` has a :t:`match arm guard`, then the
         :t:`match arm guard` is evaluated.

      #. :dp:`fls_Hw6CkR3mVzQn`
         If the :t:`match arm guard` evaluates to ``true``, then the
         :t:`match arm matcher` succeeds and its :t:`evaluation` is complete.

      #. :dp:`fls_gfD9XwVj5hab`
         If the :t:`match arm guard` evaluates to ``false``, then evaluation
         leaves the :t:`drop scope` of the related :t:`match arm` and continues
         with the next :t:`pattern`, including later
         :t:`[pattern-without-alternation]s` of the same :t:`or-pattern`.

#. :dp:`fls_Zm9VxF5pQwLd`
   If no considered :t:`pattern` causes the :t:`match arm matcher` to succeed,
   then the :t:`match arm matcher` fails.
```

Target: `src/changelog.rst`, Rust 1.95.0 entry for `Stabilize if let guards on match arms`. Please treat this as an accounting checklist rather than final paragraph-ID text. Once the source shape is settled, replace the current narrow accounting with a consolidated entry that reflects the actual landed edits.

A useful accounting shape would be:

```rst
  - Changed syntax: :s:`MatchArmGuard`

  - New syntax:

    - :s:`MatchArmGuardConditions`
    - :s:`MatchArmGuardChain`
    - :s:`MatchArmGuardCondition`
    - :s:`MatchArmGuardConditionAfterLet`
    - :s:`MatchArmGuardConditionOperand`
    - :s:`MatchArmGuardConditionOperandBeforeLet`
    - :s:`MatchArmGuardLetPattern`
    - :s:`MatchArmGuardOperand`
    - :s:`MatchArmGuardScrutinee`

  - New glossary entries:

    - :t:`match arm guard chain`
    - :t:`match arm guard condition`
    - :t:`match arm guard condition operand`
    - :t:`match arm guard let pattern`
    - :t:`match arm guard operand`
    - :t:`match arm guard scrutinee`

  - Updated glossary entries:

    - :t:`let binding`

  - Changed existing paragraphs, verifying the final list against the source:

    - :p:`fls_72JHo343O7jp`
    - :p:`fls_6bwTtGKb7ba7`
    - :p:`fls_bzhz5wjd90ii`
    - :p:`fls_st9onPgDrc8y`
    - :p:`fls_4dv7x9nh2h4e`
    - :p:`fls_sbtx1l6n2tp2`
    - :p:`fls_e02um1gb89d0`
    - :p:`fls_lbsfhg42yiqy`
    - :p:`fls_fnvr5w2wzxns`
    - :p:`fls_ptk6yibqyfzi`
    - :p:`fls_a5tmilqxdb6f`

  - Removed paragraphs, if the matcher algorithm is rewritten as above:

    - :p:`fls_k7kliy101m0f`
    - :p:`fls_k68zkb6jv0vz`
    - :p:`fls_gbb6wbmher5z`
    - :p:`fls_jl4av757yx8j`
    - :p:`fls_wkh5wztauwhu`
    - :p:`fls_f5f0x8jstp1g`
    - :p:`fls_yk8l9zjh7i0d`

  - New paragraphs for the landed definitions and rules, including:

    - the match-arm-guard-chain, operand, condition operand, and scrutinee definitions;
    - the syntax-subset definitions and exclusion-list items;
    - the bool-typing and guard-let expected-type rules;
    - the guard-let scrutinee inference rule;
    - the match-guard dynamic-semantics substeps;
    - the candidate-pattern/or-pattern retry rule;
    - the shared source-scope, drop-scope, temporary-scope, and drop-order updates;
    - the delayed copy/move rule for arm-pattern bindings with guards.
```

We should not keep the `Removed paragraph` bullet for `fls_sbtx1l6n2tp2` if this paragraph is restored and changed. We should not include `fls_uCDQMkWx5OMS` if that PR-local scope paragraph is removed before merge. If we end up adding the optional `src/patterns.rst` note qualifying `fls_pxvtqxke1enp`, we can include that new paragraph ID in the final changelog accounting too.

src/expressions.rst
PLeVasseur: I think the scope/destructor pieces need to be updated in the shared chapters, not only locally in `src/expressions.rst`.

The Reference update for this feature added more than match-expression grammar. It also added `names.scopes.pattern-bindings.match-guard-let`, updated the destructor text for local variables declared in pattern-matching match guards, and distinguished non-pattern-matching guard condition operands from pattern-matching guards for temporary scopes.

FLS still has the old single-guard-operand model in these shared rules:

- `src/entities-and-resolution.rst` `fls_xbnki64un70v` only covers bindings of a match arm;
- `src/ownership-and-deconstruction.rst` `fls_lbsfhg42yiqy` only covers the operand of a match arm guard;
- `src/ownership-and-deconstruction.rst` `fls_fnvr5w2wzxns` only covers bindings declared in a match expression;
- `src/ownership-and-deconstruction.rst` `fls_ptk6yibqyfzi` only covers the old guard operand temporary scope;
- `src/ownership-and-deconstruction.rst` `fls_a5tmilqxdb6f` would otherwise drop all temporaries before all non-or-pattern bindings, which does not model the guard-let drop order.

The PR-added local scope paragraph `fls_uCDQMkWx5OMS` describes real behavior, but I do not think it should remain the only normative source-scope rule.

Target: `src/patterns.rst`, Legality Rules, `fls_6bwTtGKb7ba7`. Replace the current `let binding` definition so it also covers bindings introduced by a match-arm guard `let` pattern:

```rst
:dp:`fls_6bwTtGKb7ba7`
A :dt:`let binding` is a :t:`binding` introduced by the :t:`pattern` of a
:t:`let statement`, an :t:`if let expression`, a
:t:`while let loop expression`, or a :t:`match arm guard let pattern`.
```

Target: `src/glossary.rst`, `let binding`. Replace the glossary definition with the same scope, preserving glossary order:

```rst
let binding
^^^^^^^^^^^

A :dt:`let binding` is a :t:`binding` introduced by the :t:`pattern` of a
:t:`let statement`, an :t:`if let expression`, a
:t:`while let loop expression`, or a :t:`match arm guard let pattern`.
```

Target: `src/patterns.rst`, Legality Rules, `fls_72JHo343O7jp`. Because `let binding` is broader after the previous change, refine this or-pattern restriction so it remains about grammar positions that use `PatternWithoutAlternation`:

```rst
:dp:`fls_72JHo343O7jp`
An :t:`or-pattern` shall not be used as the top-level :t:`pattern` of a
:t:`closure parameter`, a :t:`function parameter`, or a :t:`let statement`.
```

Target: `src/expressions.rst`, Legality Rules. Delete the PR-added local scope paragraph `fls_uCDQMkWx5OMS` rather than keeping it as the only source-scope rule in the match-expression section:

```rst
:dp:`fls_uCDQMkWx5OMS`
Each :t:`let binding` introduced in a :t:`match arm guard let pattern` is :t:`in scope` for the rest of the :t:`match arm guard` as well as the :t:`match arm body`.
```

Target: `src/entities-and-resolution.rst`, Binding Scope, immediately after `fls_xbnki64un70v`. Add the shared source-scope rule using the broadened `let binding` term:

```rst
:dp:`fls_Td7McP9rQxVn`
Each :t:`let binding` introduced by a :t:`match arm guard let pattern` is
:t:`in scope` within the following :t:`[match arm guard condition]s` of the
same :t:`match arm guard` and within the related :t:`match arm body`.
```

Target: `src/ownership-and-deconstruction.rst`, drop-scope nesting near `fls_lbsfhg42yiqy`. Replace the single guard-operand entry with entries for expression-valued guard operands and guard-let scrutinee expressions. Keep this in terms of existing expression and match-arm drop scopes; do not add a new non-expression drop construct for guard conditions unless the chapter is restructured more broadly:

```rst
* :dp:`fls_lbsfhg42yiqy`
  The parent :t:`drop scope` of a :t:`match arm guard operand` or a
  :t:`match arm guard condition operand` is the :t:`drop scope` of the
  :t:`match arm` that contains the :t:`match arm guard`.

* :dp:`fls_Jp4NsK8vYqRc`
  The parent :t:`drop scope` of the :t:`match arm guard scrutinee` of a
  :t:`match arm guard let pattern` is the :t:`drop scope` of the
  :t:`match arm` that contains the :t:`match arm guard`.
```

Target: `src/ownership-and-deconstruction.rst`, local-variable drop scope at `fls_fnvr5w2wzxns`. Replace with a rule that covers both arm-pattern bindings and pattern-matching guard `let` bindings:

```rst
:dp:`fls_fnvr5w2wzxns`
A :t:`binding` introduced by the :t:`pattern` of a
:t:`match arm matcher`, or a :t:`let binding` introduced by a
:t:`match arm guard let pattern`, is associated with the :t:`drop scope` of the
related :t:`match arm`.
```

Target: `src/ownership-and-deconstruction.rst`, temporary-scope list near `fls_ptk6yibqyfzi`. Replace the single guard-operand entry with separate temporary scopes. Expression-valued guard operands keep the old short temporary scope. Temporaries created while evaluating guard-let scrutinees use the containing match arm's drop scope, aligning with the Rust Reference's distinction between a non-pattern-matching match guard condition operand and the pattern-matching guard, if present, and body expression for a match arm. The existing match-arm operand temporary-scope item continues to cover temporaries created by the arm body itself; the new item is only for temporaries created while evaluating guard-let scrutinees.

```rst
* :dp:`fls_ptk6yibqyfzi`
  The :t:`drop scope` of a :t:`match arm guard operand` or a
  :t:`match arm guard condition operand`.

* :dp:`fls_Vq8LdN2xWmTs`
  The :t:`drop scope` of the :t:`match arm` that contains the
  :t:`match arm guard`, when the :t:`temporary` is created while evaluating a
  :t:`match arm guard scrutinee` of a :t:`match arm guard let pattern`.
```

Target: `src/ownership-and-deconstruction.rst`, Drop Order, in the list introduced by `fls_jwofws3022ar`. Add the guard-specific ordering item after the or-pattern item `fls_W2S2FrkuedYC` and before the general `Temporaries <temporary>` item `fls_a5tmilqxdb6f`, then narrow `fls_a5tmilqxdb6f` to all other temporaries.

This is needed because the current generic FLS order would otherwise read as “temporaries before bindings” for the whole match-arm drop scope. That does not model pattern-matching match guards, where evaluated guard-let conditions are processed in reverse source order and each condition's bindings are dropped before temporaries created for that condition.

```rst
* :dp:`fls_Et6PqW9nRcVm`
  Evaluated :t:`[match arm guard let pattern]s` that have
  :t:`[let binding]s` or :t:`[temporary]s` associated with the
  :t:`drop scope` of a :t:`match arm` are processed in reverse source order.
  For each such pattern, its :t:`[let binding]s`, if any, are dropped in reverse
  declaration order before the :t:`[temporary]s` created while evaluating its
  :t:`match arm guard scrutinee` are dropped in reverse creation order.

* :dp:`fls_7yybrl2J37Et`
  When a :t:`match arm guard` evaluates to ``false``, the
  :t:`[let binding]s` and :t:`[temporary]s` associated with the related
  :t:`match arm` are :t:`dropped` before the next :t:`pattern` or
  :t:`match arm` is considered.

* :dp:`fls_a5tmilqxdb6f`
  All other :t:`[temporary]s` are :t:`dropped` in reverse creation order.
```

The consolidated changelog entry for this release note should account for these pattern, glossary, scope, temporary-scope, and drop-order updates, including the changed generic temporary drop-order paragraph `fls_a5tmilqxdb6f` and the failed-guard cleanup paragraph if added. It should not include `fls_uCDQMkWx5OMS`, because that paragraph is removed before merge rather than becoming part of the final spec delta.

src/expressions.rst
PLeVasseur: This still seems to miss the Reference's guard-borrow rule for bindings introduced by the match arm pattern.

Before a match guard is evaluated, arm-pattern bindings are accessed through shared references to the matched parts of the scrutinee. Only if the guard succeeds are values moved or copied into by-value arm-pattern bindings. If the guard fails, those moves/copies do not happen.

Without that guard-specific rule, the generic by-value pattern rule in `src/patterns.rst` `fls_pxvtqxke1enp` still reads as though a by-value arm-pattern binding may move during pattern matching before guard evaluation. That conflicts with Reference `expr.match.guard.shared-ref`, `expr.match.guard.value`, and `expr.match.guard.no-mutation`, and it also explains the existing FLS immutability rule `fls_RPMOAaZ6lflI`.

Support:

- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/guard-mutability-2.rs`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/move-guard-if-let.rs`
- rustc test `tests/ui/rfcs/rfc-2294-if-let-guard/move-guard-if-let-chain.rs`

Add this guard-specific rule in `src/expressions.rst`, `Match Expressions` Dynamic Semantics, immediately after the match-guard evaluation rule `fls_sbtx1l6n2tp2`. This ties back to the existing immutability rule `fls_RPMOAaZ6lflI` and qualifies the generic by-value binding rule `fls_pxvtqxke1enp`; the copy/move timing itself should not be placed in Legality Rules:

```rst
:dp:`fls_Gc5RpT8nYvQm`
Before a :t:`match arm guard` is evaluated, each :t:`binding` introduced by the
:t:`pattern` of the related :t:`match arm matcher` is accessed in the
:t:`match arm guard` through a shared reference to the matched part of the
:t:`subject expression`'s :t:`value`.

:dp:`fls_Nw2KxL7qVmRs`
If the :t:`binding mode` of such a :t:`binding` is :t:`by value`, then the copy
or move described by :p:`fls_pxvtqxke1enp` is performed only if the related
:t:`match arm guard` evaluates to ``true``.

:dp:`fls_Rh9TsD4vQpLm`
If the related :t:`match arm guard` evaluates to ``false``, then no copy or
move into such a :t:`binding` is performed.
```

As an alternative to changing the generic rule's base wording, the qualification can also live next to the generic pattern rule. In that case, add this note to `src/patterns.rst` immediately after `fls_pxvtqxke1enp`:

```rst
:dp:`fls_<new-id>`
For a :t:`binding` introduced by the :t:`pattern` of a
:t:`match arm matcher` with a related :t:`match arm guard`, the timing of a
copy or move into a :t:`by value` binding is specified by the
:t:`match arm guard` dynamic semantics.
```

The consolidated changelog entry for this release note should include the `src/expressions.rst` IDs above. If the adjacent-note form is used, also include the new `src/patterns.rst` paragraph ID.

PLeVasseur (changes requested):
## Cross-file requirements

I think this PR needs semantic updates beyond the local `src/expressions.rst` additions. The inline suggestions below are meant to provide concrete source shapes, not to require exact category names or generated paragraph IDs.

The main gaps I see are:

- separate modeling for ordinary match-arm guard operands, expression-valued guard-chain condition operands, and guard-let scrutinees, including legality and inference;
- chain-aware dynamic semantics, including left-to-right evaluation, short-circuiting, successful guard-let binding introduction, and retrying later `|` alternatives in the same arm;
- the Reference guard-borrow rule that delays copy/move into arm-pattern bindings until the guard succeeds;
- shared source-scope, drop-scope, temporary-scope, and drop-order updates for guard-let bindings and temporaries;
- changelog accounting once the final source shape is settled.

One FLS Rust 2021 scope caveat: the current Rust Reference grammar includes attributes before guard-let patterns, but stable `rustc +1.95.0 --edition=2021` rejects that position with `E0658`. I think FLS should omit that attribute position for now, as the local branch does, or explicitly document the stable-rust divergence if the grammar is shaped around the Reference.

The Reference updates for this feature touch `match-expr`, `names/scopes`, and `destructors`, so I do not think the FLS update is complete if it only changes the local match-expression grammar and a few local legality paragraphs.

## Seeing the whole picture

There's also a feature branch on my FLS fork which has three commits showing these changes and how they land:
https://github.com/PLeVasseur/fls/tree/if-let-guards-review-fixes

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/693/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..eba53099764a6a3411e3aadf1cbb9f8496c81300)*

## Wow, this sure is detailed!

Yeah, I get it ;-;

From what I gathered, @kirtchev-adacore and @tshepang wanted more detailed and proposal style reviews.

src/expressions.rst
kirtchev-adacore: Vis-a-vis the grammar, how about the more compact
```
MatchArmGuard ::=
    $$if$$ MatchArmGuardConditions

MatchArmGuardConditions ::=
    MatchArmGuardChain
  | MatchArmGuardOperand

MatchArmGuardOperand ::=
    Operand

MatchArmGuardChain ::=
    MatchArmGuardCondition ($$&&$$ MatchArmGuardCondition)*

MatchArmGuardCondition ::=
    MatchArmGuardConditionOperand
  | MatchArmGuardLetPattern

MatchArmGuardLetPattern ::=
    $$let$$ Pattern $$=$$ MatchArmGuardScrutinee
```
plus the syntactic clarifications about `MatchArmGuardConditionOperand` and `MatchArmGuardScrutinee` (there is no need to add productions for those).

I could not comprehend the roles of `MatchArmGuardConditionOperandBeforeLet` and `MatchArmGuardConditionAfterLet`, so I left them out.

kirtchev-adacore:
```
:dp:`fls_DT4N2rr6wpvZ`
A :t:`match arm guard chain` is a sequence of one or more
:t:`[match arm guard condition]s`, separated by ``&&``, that contains at least
one :t:`match arm guard let pattern`.
```

I am not sure where you got the "contains at least one match arm guard let pattern". I do not see such a requirement in the Reference.

kirtchev-adacore: I am not sure what the purpose of the following rules is
```
:dp:`fls_Uv7QsP2nKcDx`
A :t:`match arm guard operand` is the :t:`expression` in category
:s:`MatchArmGuardOperand` of a :t:`match arm guard` that does not have a
:t:`match arm guard chain`.

:dp:`fls_Er5TmL9bXwVq`
A :t:`match arm guard condition operand` is the :t:`expression` in category
:s:`MatchArmGuardConditionOperand` of a :t:`match arm guard condition`.

:dp:`fls_Yq4NhC8sRpJm`
A :t:`match arm guard scrutinee` is the :t:`expression` in category
:s:`MatchArmGuardScrutinee` of a :t:`match arm guard let pattern`.
```
This is like saying "foo bar is a FooBar". If you want to give a term to a construct, then you should phrase the rules similar to
```
A match arm guard operand is a construct which/that ...
```

src/expressions.rst
kirtchev-adacore: Pete's suggestions LGTM!

src/expressions.rst
kirtchev-adacore: Pete's suggestions LGTM!

src/expressions.rst
kirtchev-adacore: Agreed. I am retracting my suggestion.

tshepang:
made obsolete by #709

PR #694, opened 2026-04-15
document missing import restrictions

tshepang:


src/entities-and-resolution.rst
```diff
@@ -1164,8 +1176,6 @@ The following is a selective import. The imported functions are
    use outer_module::inner_module
        {crate_visible_function, visible_function}
 
-.. rubric:: Legality Rules
-
```
jyn514, line 1168: Was this deletion intentional?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/f5c48f51906cc12b5745ced6658d96ea0850552a..a073e3e3b1931a0300a8fd8e5e31198839ce3f40)*

tshepang: yeah, it was stray text... it's in a separate commit to make that clear

jyn514 (approved):

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1135,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
 If two :t:`[glob import]s` import the same :t:`entity` under the same :t:`name`,
 the :t:`visibility` of the :t:`name` is the most permissive one.
 
+:dp:`fls_sUhnfV62HJrb`
+When using :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` to import the current :t:`crate`, you must use a :t:`renaming` to define the :t:`binding` name.
```
kirtchev-adacore: Could you point me to the Reference PR or text where you took these rules from?

Regardless, since the FLS is not a tutorial, you should avoid expressions such as "you must" (who is "you"? the programmer, the author of the conforming tool, the assessor?).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/f5c48f51906cc12b5745ced6658d96ea0850552a..a073e3e3b1931a0300a8fd8e5e31198839ce3f40)*

tshepang: see https://doc.rust-lang.org/nightly/reference/items/use-declarations.html#restrictions

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1135,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
 If two :t:`[glob import]s` import the same :t:`entity` under the same :t:`name`,
 the :t:`visibility` of the :t:`name` is the most permissive one.
 
+:dp:`fls_sUhnfV62HJrb`
+When :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` is used to import the current :t:`crate`, a :t:`renaming` must be used to define the :t:`binding` name.
```
PLeVasseur: This still looks incomplete relative to the upstream `self`-import change. Earlier FLS paragraphs still encode the pre rust-lang/rust#146972 behavior: `fls_cw006jhlboa` still requires a single-segment `self` use-path to appear only inside a nesting import, which rejects `use self as name;`, while `fls_RUiFQ17bmRLt` still allows unrenamed `use self::{self};` when the import path prefix is non-empty.

After rust-lang/rust#146972, those move together: renamed current-module imports such as `use self as name;` are allowed, and unrenamed current-module imports like `use self::{self};` are rejected.

Could this PR also update those earlier `self` paragraphs so the new restriction block and the older legality rules stay aligned? A minimal fix would be to loosen `fls_cw006jhlboa` for the renamed single-segment case and tighten `fls_RUiFQ17bmRLt` so the non-empty-prefix exception does not keep `use self::{self};` valid.

Support:
- rust-lang/rust#146972
- Reference `items.use.restrictions.self-alias`
- Reference rust-lang/reference#2136

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1135,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
+
+:dp:`fls_QGdeRTe0H1Uc`
+When :t:`keyword` ``super`` is used to import a parent :t:`module`, a :t:`renaming` must be used to define the :t:`binding` name.
```
PLeVasseur: This new `super` rule still conflicts with the earlier path-legality rule. `fls_7k88ypcgaoff` currently allows a `super` segment only as the first segment or after another `super`, so the FLS still rejects `self::super` and `self::super::super`. Upstream now allows `super` after an initial `self`, and the Reference examples for this restriction explicitly include `use self::super as parent3;`.

Could this PR also relax `fls_7k88ypcgaoff` so `super` may follow an initial `self`, or otherwise carve out the same allowance for `use` paths, so the earlier path rules do not keep rejecting forms this paragraph is meant to document?

Support:
- rust-lang/rust#146972
- Reference `paths.qualifiers.super.allowed-positions`
- Reference `paths.qualifiers.super.repetition`
- Reference `items.use.restrictions.super-alias`
- Reference rust-lang/reference#2136

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1135,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
+
+:dp:`fls_aam34hsRmKU2`
+An :t:`external prelude` cannot be imported.
```
PLeVasseur: This sentence reads too broadly. Rust 2021 still allows imports through extern-prelude crate names, such as `use ::std::io;`. The disallowed case is importing bare leading `::` itself, for example `use ::{self as root};`.

As written, `An external prelude cannot be imported.` reads as though extern-prelude imports are disallowed wholesale.

Could this be restated in terms of `::` as the extern-prelude root instead? If so, the earlier global-path rules probably also need a follow-on update so Rust 2021 no longer appears to permit `::self`, `::super`, `::crate`, or `::$crate`.

Support:
- rust-lang/rust#146972
- Reference `items.use.restrictions.extern-prelude`
- Reference `paths.qualifiers.global-root.edition2018`
- Reference rust-lang/reference#2136

src/entities-and-resolution.rst
PLeVasseur: `$crate` probably should not be folded into the same sentence as `crate` here. Earlier FLS text treats it as macro-specific in two ways: `fls_774uryecc2sx` only permits it within a macro transcriber, and `fls_yrpem8vhxpr5` resolves it to the crate that defines the expanded macro. Grouping it under "used to import the current crate" makes `$crate` read like an ordinary spelling of `crate`, which is weaker than the current FLS model and the Reference's separate `items.use.restrictions.macro-crate-alias` rule.

Could this be split into separate `crate` and `$crate` restrictions, or reworded so the `$crate` half matches the existing FLS model and the Reference's separate `items.use.restrictions.macro-crate-alias` rule?

Support:
- rust-lang/rust#146972
- Reference `items.use.restrictions.macro-crate-alias`
- Reference `paths.qualifiers.macro-crate`
- Reference rust-lang/reference#2136

src/changelog.rst
```diff
@@ -28,6 +28,14 @@ Language changes in Rust 1.95.0
   - Lints are outside the scope of the FLS.
 
 - `Support importing path-segment keywords with renaming <https://github.com/rust-lang/rust/pull/146972>`_
+
+  - New paragraphs:
+
+    - :p:`fls_sUhnfV62HJrb`
+    - :p:`fls_QGdeRTe0H1Uc`
+    - :p:`fls_aam34hsRmKU2`
+    - :p:`fls_LV94x3HlpBWk`
```
PLeVasseur: I rechecked the pre rust-lang/reference#2136 Reference snapshot, and the enum-variant-through-type-alias restriction was already present there as `items.use.restrictions.variant`. Since rust-lang/rust#146972 is specifically about keyword imports, adding `fls_LV94x3HlpBWk` under this Rust 1.95 language-change entry makes this look like an FLS correction rather than part of the 1.95 change itself.

Could this paragraph ID move out of the `Support importing path-segment keywords with renaming` entry and into the existing `FLS corrections` section instead, unless there is a stronger reason to tie it to rust-lang/rust#146972?

Support:
- Reference snapshot before rust-lang/reference#2136 already contains `items.use.restrictions.variant` (`rust-lang/reference@36016fc5164d3e10d0288e71f57c88dca937bd80`)
- rust-lang/reference#2136
- rust-lang/rust#146972

PLeVasseur (changes requested):
These comments are blocking because the PR currently adds only part of the upstream keyword-import model. In particular, the existing `self` and `super` legality rules still conflict with the new restriction paragraphs, so the FLS would remain internally inconsistent after merge. The remaining comments cover precision of the extern-prelude wording, preservation of the `$crate` distinction, and changelog classification.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/f5c48f51906cc12b5745ced6658d96ea0850552a..0b31ecc72d8cfe10af1bb4c443537eabccbeaad8)*

src/entities-and-resolution.rst
kirtchev-adacore: How about imitating the wording of 14.7:25:

A :t:`simple import` where the first :t:`path segment` of its :t:`simple path` is keyword `crate` shall be subject to a :t:`renaming`.

src/entities-and-resolution.rst
kirtchev-adacore: How about the following wording:

> A :t:`simple import` where the first :t:`path segment` of its :t:`simple path` is either :t:`keyword` `crate` or :t:`keyword` `$crate` shall be subject to a :t:`renaming`.

src/entities-and-resolution.rst
kirtchev-adacore: How about the following wording:
> A :t:`simple import` where any :t:`path segment` of its :t:`simple path` denotes :t:`keyword` `super` shall be subject to a :t:`renaming`.

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1135,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
+
+:dp:`fls_LV94x3HlpBWk`
+A :t:`simple import` cannot refer to :t:`[enum variant]s` through a :t:`type alias`.
```
kirtchev-adacore:
```suggestion
A :t:`simple import` shall not refer to :t:`[enum variant]s` through a :t:`type alias`.
```
Remember that this is a requirement aimed at the programmer, not a fact of the conforming tool. Hence you must use normative language.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/f5c48f51906cc12b5745ced6658d96ea0850552a..0b31ecc72d8cfe10af1bb4c443537eabccbeaad8)*

kirtchev-adacore (changes requested):

src/entities-and-resolution.rst
tshepang: https://github.com/rust-lang/fls/pull/694/commits/0b31ecc72d8cfe10af1bb4c443537eabccbeaad8

tshepang: > A :t:`simple import` where the first :t:`path segment` of its :t:`simple path` is keyword `crate` shall be subject to a :t:`renaming`.

went with 752f06e, since your suggestions would not fit, because
- the rules apply also to nested paths
- the rules apply also to cases where the first segment is not the keyword

kirtchev-adacore: How about changing 14.7:25 into

The :t:`simple path` of a :t:`nested import` or a :t:`simple import` where the last :t:`path segment` is :t:`keyword` ``crate``, :t:`keyword` ``$crate``, :t:`keyword` ``self``, or :t:`keyword` ``super`` shall be subject to a :t:`renaming`.

src/entities-and-resolution.rst
kirtchev-adacore: I made a blanket suggestion in https://github.com/rust-lang/fls/pull/694#discussion_r3217526038.

src/entities-and-resolution.rst
kirtchev-adacore: I made a blanket suggestion in https://github.com/rust-lang/fls/pull/694#discussion_r3217526038.

src/entities-and-resolution.rst
kirtchev-adacore: How about this wording:

A :t:`global path` shall not consist of a single :t:`path segment` expressed as :t:`keyword` ``crate``, :t:`keyword` ``$crate``, :t:`keyword` ``self``, or :t:`keyword` ``super``.

**This comment is no longer relevant.**

src/entities-and-resolution.rst
tshepang: - `crate` (and `$crate`) is only allowed at the beginning of an import path
- the remaining keywords, `self` and `super`, have different rules (where the former can be used without a renaming, depending)

src/entities-and-resolution.rst
tshepang: > Could this PR also update those earlier `self` paragraphs so the new restriction block and the older legality rules stay aligned? A minimal fix would be to loosen `fls_cw006jhlboa` for the renamed single-segment case and tighten `fls_RUiFQ17bmRLt` so the non-empty-prefix exception does not keep `use self::{self};` valid.

am thinking of:
- removing `fls_cw006jhlboa` because I don't quite understand it... somehow feels redundant
- separating `fls_RUiFQ17bmRLt` into 2 separate rules, because "single path segment" is confusing

src/entities-and-resolution.rst
tshepang: @PLeVasseur `fls_7k88ypcgaoff` looks like it was wrong long before this PR, and I even think we should remove it... what do you think

src/entities-and-resolution.rst
tshepang: > Could this be restated in terms of `::` as the extern-prelude root instead? If so, the earlier global-path rules probably also need a follow-on update so Rust 2021 no longer appears to permit `::self`, `::super`, `::crate`, or `::$crate`.

@PLeVasseur can you provide alternative wording... it's not clear to me what an extern prelude is (Reference does not make it clear either), and I also don't see how any rule appears to permit the import paths shown

src/entities-and-resolution.rst
tshepang: I actually don't see any ambiguity here... both are indeed used to import the current crate, and any additional rules are written elsewhere (like the fact that `$crate` can only be used in macros), meaning that I find Reference text redundant.

Leaving it this bare makes it less confusing, in that if you say "When using `$crate` in a macro transcriber to import the current crate, you must use a renaming" makes it look like this is implied, "When using `$crate` in outside of macro transcriber to import the current crate, you need not use a renaming".

If it still feels dirty mixing them, I can create a separate paragraph, but that will have the exact same text, other than `crate` and `$crate`.

src/changelog.rst
tshepang: 9c70941741b6247a77d912021c7211d56eca1401

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1133,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
 If two :t:`[glob import]s` import the same :t:`entity` under the same :t:`name`,
 the :t:`visibility` of the :t:`name` is the most permissive one.
 
+:dp:`fls_sUhnfV62HJrb`
+When :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` is used to import the current :t:`crate`, the keyword shall be subject to a :t:`renaming`.
```
kirtchev-adacore:
```suggestion
When a :t:`simple path segment` expressed as :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` is used to import the current :t:`crate`, the :t:`simple import` shall be subject to a :t:`renaming`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..9c70941741b6247a77d912021c7211d56eca1401)*

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1133,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
+
+:dp:`fls_QGdeRTe0H1Uc`
+When :t:`keyword` ``super`` is used to import a parent :t:`module`, the keyword shall be subject to a :t:`renaming`.
```
kirtchev-adacore:
```suggestion
When a :t:`simple path segment` expressed as :t:`keyword` ``super`` is used to import a parent :t:`module`, the :t:`simple import` shall be subject to a :t:`renaming`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..9c70941741b6247a77d912021c7211d56eca1401)*

src/entities-and-resolution.rst
kirtchev-adacore: How about
```
A :t:`simple path` consisting of namespace qualifier `::`, followed by an optional :t:`simple path
segment` expressed as :t:`keyword` `self`, shall not be used to import an :t:`external prelude`.
```

tshepang: *__update__: comment obsolete*

@kirtchev-adacore does 215f85e198ad70bcfe9236faf609f8a2ee6ff31c work
I removed mention of extern prelude since that makes it look like the import can be used for other purposes

src/entities-and-resolution.rst
tshepang: 276507752bc4836ee64d660139fb6f2ae4b72954 and d7e8007424593264405ba965ac3e914e7f1fec66

kirtchev-adacore: I am not seeing the change for `super` though, the latest version of the MR still reflects the "keyword shall be subject to a renaming phrasing".

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1133,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
 the :t:`visibility` of the :t:`name` is the most permissive one.
 
+:dp:`fls_sUhnfV62HJrb`
+When a :t:`path segment` expressed as :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` is used to import the current :t:`crate`, the :t:`path segment` shall be subject to a :t:`renaming`.
+
+:dp:`fls_QGdeRTe0H1Uc`
+When :t:`keyword` ``super`` is used to import a parent :t:`module`, the keyword shall be subject to a :t:`renaming`.
+
+:dp:`fls_aam34hsRmKU2`
+A :t:`simple path` consisting of namespace qualifier ``::``, followed by a :t:`path segment` expressed as :t:`keyword` ``self``, shall not be used.
```
kirtchev-adacore:
```suggestion
A :t:`simple path` consisting of namespace qualifier ``::`` followed by a :t:`path segment` expressed as :t:`keyword` ``self`` shall not be used.
```
I had suggested the commas when I proposed that `self` is optional, but it is not, so the commas are not needed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..276507752bc4836ee64d660139fb6f2ae4b72954)*

src/entities-and-resolution.rst
tshepang: ah, missed that... 857cd5e

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1133,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
 If two :t:`[glob import]s` import the same :t:`entity` under the same :t:`name`,
 the :t:`visibility` of the :t:`name` is the most permissive one.
 
+:dp:`fls_sUhnfV62HJrb`
+When a :t:`path segment` expressed as :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` is used to import the current :t:`crate`, the :t:`path segment` shall be subject to a :t:`renaming`.
+
+:dp:`fls_QGdeRTe0H1Uc`
+When a :t:`path segment` expressed as :t:`keyword` ``super`` is used to import a parent :t:`module`, the :t:`path segment` shall be subject to a :t:`renaming`.
```
kirtchev-adacore: These are semantically incorrect. A path segment cannot be subject to a renaming, it is the simple import that is. The grammar is:
```
SimpleImport ::=
    SimplePath Renaming?
```
The renaming applies to the whole simple import, not to a particular path segment. Also look at my original suggestion:

When a :t:`simple path segment` expressed as :t:`keyword` ``super`` is used to import a parent :t:`module`, the **:t:`simple import` shall be subject to a :t:`renaming`**.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..29ca7f81ca4d9b9ff38d8a2e02309f1daa553835)*

kirtchev-adacore (changes requested):

src/entities-and-resolution.rst
kirtchev-adacore: After a lengthy discussion with Tshepang on Meet, we determined that
1. We may have a serious problem with out [Path](https://rust-lang.github.io/fls/entities-and-resolution.html#paths) syntax.
2. Assuming that we fix our Path syntax, Tshepang's original wording (**path segment** shall be subject to ...) is actually correct.

Tagging @traviscross and @ehuss to confirm / disprove.

kirtchev-adacore: A few clarifications:
1. We do not have a serious problem with our Path syntax, I simply misread the grammar.
2. Based on the [discussion](https://hackmd.io/kc8zqym5Tla_o76fmXZJvw?view) on renamings, I propose the following wording:

When a :t:`path segment` expressed as :t:`keyword` ``super`` is used to import a parent :t:`module`, the imported :t:`entity` shall be subject to a :t:`renaming`.

Ditto for `crate` and `$crate`.

PLeVasseur: I think Hristian's latest wording matches how I understand the meeting conclusion: the thing being renamed is the :t:`entity` imported by the related :t:`simple import`, not the textual :t:`path segment`. This is the FLS semantic model even though the grammar attaches `Renaming?` to `SimpleImport`. It also fits the existing FLS rule in `fls_FILuR3pfwjw3`, which already says an entity imported by a simple import subject to a renaming is brought into scope under the renamed name.

I would suggest replacing the current `crate`/`$crate` and `super` paragraphs along these lines. I would also suggest keeping `$crate` distinct enough to preserve `fls_yrpem8vhxpr5`, where `$crate` resolves to the crate that declares the macro being expanded. The Reference uses "current crate" in the macro-transcriber context, while the FLS already has a more precise macro-definition-crate model. The `self` restriction is covered in the separate self-import thread.

```rst
:dp:`fls_sUhnfV62HJrb`
When a :t:`path segment` expressed as :t:`keyword` ``crate`` is used to import
the current :t:`crate`, the :t:`entity` imported by the related
:t:`simple import` shall be subject to a :t:`renaming`.

:dp:`fls_61qKMG5tuv12`
When a :t:`path segment` expressed as :t:`keyword` ``$crate`` is used within a
:t:`macro transcriber` to import the :t:`crate` that declares the :t:`macro`
that is being expanded, the :t:`entity` imported by the related
:t:`simple import` shall be subject to a :t:`renaming`.

:dp:`fls_QGdeRTe0H1Uc`
When a :t:`path segment` expressed as :t:`keyword` ``super`` is used to import a
parent :t:`module`, the :t:`entity` imported by the related :t:`simple import`
shall be subject to a :t:`renaming`.
```

Use separate `crate` and `$crate` restriction paragraphs. That preserves the existing FLS `$crate` resolution rule in `fls_yrpem8vhxpr5` and maps directly to the Reference anchors `items.use.restrictions.crate-alias` and `items.use.restrictions.macro-crate-alias`. The Rust 1.95 keyword-import changelog entry should list `fls_sUhnfV62HJrb`, `fls_QGdeRTe0H1Uc`, and `fls_61qKMG5tuv12` as new paragraphs relative to the merge base.

src/entities-and-resolution.rst
PLeVasseur: For `fls_cw006jhlboa`, I think the cleanest route is to delete it and replace `fls_RUiFQ17bmRLt` with a semantic current-module import rule rather than a rule limited to a single-segment `self` simple path.

The cases I think we need to preserve are:

1. rust-lang/rust#146972 allows renamed current-module imports: `use {self as this_module};`, `use self as this_module2;`, `use self::{self as this_module3};`, and `use {self::self as this_module4};`.
2. The same change still rejects unrenamed current-module imports: `use {self};`, `use self;`, `use self::{self};`, and `use {self::self};`.
3. The post-rust-lang/rust#146972 Rust 2018-and-later use-path model keeps braced parent-entity imports valid in Rust 2021, including `use a::b::{self};`, `use a::b::{self as name};`, `use a::{b::self};`, and `use a::{b::self as name};`. Direct final-`self` paths such as `use a::b::self as name;` are still rejected by Rust 1.95.0 with E0429; the separate path-position comment updates `fls_opn5n5t2mo3m` so the final-`self` exception is limited to the braced/nesting-import form.

As written, `fls_cw006jhlboa` blocks the first group because it still requires a single-segment `self` use path to be part of a nesting import. As written, `fls_RUiFQ17bmRLt` leaves `use self::{self};` looking valid because it has a non-empty import path prefix, and a single-segment-only replacement would still miss `use {self::self};`.

I would delete `fls_cw006jhlboa`. The replacement below keeps the current-module `self` import restriction where it belongs: on imports of the current module, not on every final-`self` parent-entity import. This does not affect the separate `use self::*;` glob-import restriction.

After deleting `fls_cw006jhlboa`, a copy-paste replacement for `fls_RUiFQ17bmRLt` could be as follows:

```rst
:dp:`fls_RUiFQ17bmRLt`
When a :t:`path segment` expressed as :t:`keyword` ``self`` is used to import
the current :t:`module`, the :t:`entity` imported by the related
:t:`simple import` shall be subject to a :t:`renaming`.
```

That keeps `use a::b::{self};` and `use a::{b::self};` valid, because the `self` there imports the parent entity `a::b` under its own name. It rejects `use self::{self};` and `use {self::self};`, because those import the current module without a renaming. This requires the separate `crate` and `super` restriction paragraphs to use the imported-entity wording from the renaming thread; those paragraphs then reject unrenamed cases such as `use crate::{self};` and `use super::{self};`, while still allowing parent-entity imports such as `use crate::{m::self};` from a scope where that does not duplicate an existing name.

As a quick check with `rustc +1.95.0 --edition=2021`, `use {self as this_module};`, `use self as this_module2;`, `use self::{self as this_module3};`, and `use {self::self as this_module4};` are valid, while `use {self};`, `use self;`, `use self::{self};`, and `use {self::self};` remain invalid.

The Reference examples under `items.use.restrictions.self-alias` are a useful test matrix for this rule.

List `fls_cw006jhlboa`, `fls_Pxc0Ts8Y7pfW`, and `fls_hv3xT2CjZuxc` as removed, and keep `fls_RUiFQ17bmRLt` listed as changed. The corresponding simple-import semantics in `fls_wRmvtgQkFA6w` also need the update from the simple-import thread so final-`self` nesting imports bring the parent entity selected by the import prefix into scope.

src/entities-and-resolution.rst
PLeVasseur: I think I would keep a revised `fls_7k88ypcgaoff` rather than remove it outright, because we still need a general path-legality rule that rejects non-leading `super` in paths such as `foo::super`, `crate::super`, and `$crate::super`; rust-lang/rust#146972 added the `self::super` allowance that now appears in the Reference path-qualifier rules.

The rustc implementation added this exact prefix check in `build_reduced_graph_for_use_tree`: the prefix before a `super` segment may contain `super` segments, plus `self` at position 0. The updated diagnostic says `super` can be used "in start position, after `self`, or after another `super`." rust-lang/reference#2136 then added `use self::super as parent3;` to the valid `super` import examples.

A direct FLS replacement could be:

```rst
:dp:`fls_7k88ypcgaoff`
If a :t:`path segment` is expressed as :t:`keyword` ``super``, then each
:t:`path segment` that precedes it in the :t:`path` shall be expressed as
:t:`keyword` ``super``, except that the first :t:`path segment` of the
:t:`path` may be expressed as :t:`keyword` ``self``.
```

This makes the intended cases valid:

```rust
use super as parent;
use super::{self as parent2};
use self::super as parent3;
use self::super::super as grandparent3;
use super::super as grandparent;
use super::super::{self as grandparent2};
```

It still rejects the cases rustc continues to reject, such as `use foobar::super as name;`, `use self::foo::super as name;`, `use crate::super as name;`, and `use $crate::super as name;`. The unchanged first-segment rule for `crate`, `$crate`, `self`, and `Self` continues to handle those keywords separately.

This intentionally does not handle `::super`. This replacement depends on the global-path paragraph proposed in the `::` thread; that paragraph rejects `::super` for the separate Rust 2021 global-root reason. Since `fls_7k88ypcgaoff` is a general path legality paragraph, the changelog should record it as changed.

src/entities-and-resolution.rst
PLeVasseur: The distinction I am trying to preserve is not that the alias requirement is different for `$crate`; it is that the imported entity is described differently in the FLS model.

The existing FLS has two relevant rules:

- `fls_774uryecc2sx`: a path starting with `$crate` appears only within a macro transcriber.
- `fls_yrpem8vhxpr5`: `$crate` resolves to the crate that declares the macro that is being expanded.

So a combined `crate`/`$crate` paragraph that says both are used to import the current crate does not quite preserve the surrounding FLS text. In the macro case, `$crate` is intentionally definition-crate based. That distinction is also why the Reference keeps separate anchors for `items.use.restrictions.crate-alias` and `items.use.restrictions.macro-crate-alias`, even though the rules are adjacent and closely related. The Reference uses "current crate" in the macro-transcriber context, while the FLS source can preserve the more precise macro-definition-crate model it already has in `fls_yrpem8vhxpr5`.

I would suggest using the split wording from the renaming thread, with generated ID `fls_61qKMG5tuv12` for the `$crate` paragraph:

```rst
:dp:`fls_sUhnfV62HJrb`
When a :t:`path segment` expressed as :t:`keyword` ``crate`` is used to import
the current :t:`crate`, the :t:`entity` imported by the related
:t:`simple import` shall be subject to a :t:`renaming`.

:dp:`fls_61qKMG5tuv12`
When a :t:`path segment` expressed as :t:`keyword` ``$crate`` is used within a
:t:`macro transcriber` to import the :t:`crate` that declares the :t:`macro`
that is being expanded, the :t:`entity` imported by the related
:t:`simple import` shall be subject to a :t:`renaming`.
```

Use two paragraphs because they map cleanly to the two Reference anchors and avoid folding the macro-specific `$crate` rule into an ordinary `crate` sentence. List `fls_sUhnfV62HJrb` and `fls_61qKMG5tuv12` as new paragraphs in the Rust 1.95 keyword-import changelog entry; `fls_774uryecc2sx` continues to provide the placement restriction for `$crate`.

src/changelog.rst
PLeVasseur: This part is addressed now: `fls_LV94x3HlpBWk` has been moved out of the Rust 1.95 `Support importing path-segment keywords with renaming` entry and into `FLS corrections`.

I would keep the thread unresolved until the remaining source edits are applied. With the recommended `self`/`super`/`$crate`/`::`/simple-import source shape, including the Rust 1.95.0 braced final-`self` cases such as `use a::{b::self as name};`, use the following base-relative changelog accounting.

The Rust 1.95 keyword-import entry should be:

```rst
  - New paragraphs:

    - :p:`fls_sUhnfV62HJrb`
    - :p:`fls_QGdeRTe0H1Uc`
    - :p:`fls_aam34hsRmKU2`
    - :p:`fls_P6dFw89ZDKv2`
    - :p:`fls_61qKMG5tuv12`

  - Changed paragraphs:

    - :p:`fls_opn5n5t2mo3m`
    - :p:`fls_WAA4WmohGu6T`
    - :p:`fls_2bkcn83smy2y`
    - :p:`fls_iuzvtr3oax1o`
    - :p:`fls_90hQvSh7Bfyg`
    - :p:`fls_wRmvtgQkFA6w`
    - :p:`fls_kz2Gij5wHXnl`
    - :p:`fls_ar03D5rxjzy0`
    - :p:`fls_iQOgxNihUEr7`
    - :p:`fls_RUiFQ17bmRLt`
    - :p:`fls_7k88ypcgaoff`

  - Removed paragraphs:

    - :p:`fls_cw006jhlboa`
    - :p:`fls_yY58pFpkig9o`
    - :p:`fls_Pxc0Ts8Y7pfW`
    - :p:`fls_hv3xT2CjZuxc`

  - Updated glossary entries:

    - :t:`simple import`
```

I trial-applied this accounting with the source snippets from the other comments and `./make.py --clear` builds. I also expanded the Rust 1.95.0 edition-2021 test matrix to cover valid braced final-`self` imports, including `use crate::foo::{bar::foobar::quxbaz::self};`, and invalid direct/intermediate forms. This accounting assumes the `::` import-prefix preservation wording is folded into the changed `fls_WAA4WmohGu6T`; if it is split into a separate paragraph, add that new paragraph to the same changelog entry.

I do not think any further changelog change is needed for the enum-variant-through-type-alias classification itself, but please verify that `fls_LV94x3HlpBWk` remains under `FLS corrections` and that the keyword-import entry is rechecked after the source-shape edits land.

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1133,18 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
+
+:dp:`fls_aam34hsRmKU2`
+A :t:`simple path` consisting of namespace qualifier ``::`` followed by a :t:`path segment` expressed as :t:`keyword` ``self`` shall not be used.
```
PLeVasseur: This moves in the right direction compared with the earlier "external prelude cannot be imported" wording. I think it is still a bit too narrow for the surrounding FLS path model.

For Rust 2021, the Reference rule is two-part:

- Leading `::` resolves from the extern prelude and must be followed by the name of a crate.
- The `::` extern-prelude root itself cannot be imported, so `use ::{self as root};` is rejected.

That preserves valid imports through extern-prelude crate names, such as `use ::std::io;` when `std` is in the external prelude, while rejecting direct root imports. The rustc tests for rust-lang/rust#146972 also reject keyword starts after `::`, such as `::crate`, `::super`, and `::self`.

As written, the FLS text only rejects `::self`. The general FLS `global path` and keyword-segment rules should also reject `::crate`, `::super`, and `::$crate` in Rust 2021 use paths.

Since FLS documents Rust 2021, I would split this into a general global-path legality rule plus the import-root restriction. The first paragraph follows the Reference path-qualifier rule for Rust 2018 and later and should live with the general Paths legality rules near the definition of :t:`global path`. The second follows the FLS import-prefix model and should stay with the Use Imports restrictions so it directly covers `use ::{self as root};`; the bare `::` prefix there has no first path segment, so it needs a separate root-import rule.

This also depends on the import-prefix construction preserving the leading `::` namespace qualifier. The simple-import thread updates `fls_WAA4WmohGu6T` so an empty prefix with no namespace qualifier resolves to the current module, while a prefix that starts with `::` remains a global prefix. That keeps `use {self as this_module};` distinct from `use ::{self as root};` in the FLS model.

```rst
:dp:`fls_P6dFw89ZDKv2`
The first :t:`path segment` of a :t:`global path` shall be expressed as an
:t:`identifier` whose :t:`name` matches the :t:`name` of a
:t:`candidate external prelude entity`.

:dp:`fls_aam34hsRmKU2`
A :t:`simple import` whose :t:`import path prefix` consists only of
:t:`namespace qualifier` ``::`` and whose :t:`simple path` consists of a single
:t:`path segment` expressed as :t:`keyword` ``self`` shall not be used.
```

The first paragraph is what keeps `use ::std::io;` and `use ::std::{io};` valid when `std` denotes a candidate external prelude entity, while rejecting keyword starts such as `::crate`, `::super`, and `::self` in Rust 2021. The second paragraph captures the Reference's `items.use.restrictions.extern-prelude` rule for importing the bare extern-prelude root. The changelog should account for both `fls_P6dFw89ZDKv2` and `fls_aam34hsRmKU2` as new paragraphs relative to the merge base, and keep `fls_WAA4WmohGu6T` listed as changed for the import-prefix construction update.

src/entities-and-resolution.rst
```diff
@@ -1049,9 +1049,7 @@ An :dt:`import path prefix` is the fully constructed :t:`path` prefix of a
    the current :t:`use import`.
 
 :dp:`fls_2bkcn83smy2y`
-A :t:`simple import` is a :t:`use import` that brings all :t:`entities <entity>`
-it refers to into scope, optionally with a different
-:t:`name` than they are declared with by using a :t:`renaming`.
+A :dt:`simple import` is a :t:`use import` that brings a :t:`simple path` into scope, optionally with a :t:`renaming`.
```
PLeVasseur: I think this definition should stay entity/name based. A path is syntax; it does not get brought into scope. A `use` declaration brings selected entities into scope, usually by creating local name bindings, and `as` changes the local name for the imported entity. It can also use `as _`, where the imported entity is added to scope without a name.

That distinction matters for the new keyword-import restrictions because the meeting discussion resolved the renaming target as the imported entity, not the path or path segment. It also matters for final-`self` nesting imports: `use a::{b::self};` imports the parent entity selected by the constructed import prefix, not an entity resolved through a literal final `self` member.

Suggested wording for both this paragraph and the glossary entry:

```rst
A :dt:`simple import` is a :t:`use import` that brings
:t:`entities <entity>` selected by its :t:`simple import path`, or by its
:t:`import path prefix` when its :t:`simple path` ends in :t:`keyword` ``self``
and it appears in a :t:`nesting import`, into :t:`scope`, either under their
declared :t:`[name]s`, under a different :t:`name` by using a :t:`renaming`, or
without a :t:`name` by using a :t:`renaming` with character underscore ``_``.
```

I think the empty-prefix and global-prefix cases should also be explicit. Rust 1.95.0 accepts both `use {self as this_module};` and `use {self::self as this_module};`, where the prefix is local and empty. But `use ::{self as root};` is different in Rust 2021: the prefix is the extern-prelude root and is rejected by the separate `::` import restriction. I would update `fls_WAA4WmohGu6T` so empty local prefixes and global `::` prefixes remain distinct:

```rst
:dp:`fls_WAA4WmohGu6T`
An :dt:`import path prefix` is the fully constructed :t:`path` prefix of a
:t:`use import`. An :t:`import path prefix` with no :t:`[path segment]s` and no
:t:`namespace qualifier` resolves to the current :t:`module`. An
:t:`import path prefix` that starts with :t:`namespace qualifier` ``::``
preserves that :t:`namespace qualifier` in the constructed prefix. An
:t:`import path prefix` for a given :t:`simple import` or :t:`glob import` is
constructed as follows:

#. :dp:`fls_IPYvldMqduf4`
   Start the :t:`import path prefix` as follows:

   * :dp:`fls_MOXId37fcNPY`
     If the :t:`use import` is a :t:`simple import`, then start with the
     :t:`simple import`'s :t:`simple path` :t:`path prefix`.

   * :dp:`fls_2UyFcB6Our1v`
     If the :t:`use import` is a :t:`glob import`, then start with the
     :t:`glob import`'s :t:`simple path prefix`.

   * :dp:`fls_irdKqoYzBM0M`
     If the :t:`use import` is a :t:`nesting import`, then start with the
     :t:`nesting import`'s :t:`simple path prefix`.

#. :dp:`fls_gAWsqibl4GLq`
   Then if the current :t:`use import` is the child of a :t:`nesting import`,
   prepend the :t:`nesting import`'s :t:`simple path prefix` to the
   :t:`import path prefix`. Repeat this step with the :t:`nesting import` as
   the current :t:`use import`.
```

Then I would update `fls_wRmvtgQkFA6w` so final-`self` simple imports inside brace syntax import the parent entity selected by the fully constructed prefix:

```rst
:dp:`fls_wRmvtgQkFA6w`
A :t:`simple import` brings :t:`[name]s` into :t:`scope` as follows:

* :dp:`fls_kz2Gij5wHXnl`
  If the :t:`simple import` appears in a :t:`nesting import` and the last
  :t:`path segment` of its :t:`simple path` is expressed as :t:`keyword`
  ``self``, then bring the :t:`entity` in :t:`type namespace` that the
  :t:`import path prefix` resolves to into :t:`scope`.

* :dp:`fls_ar03D5rxjzy0`
  If the :t:`simple path` is :t:`keyword` ``self``, then bring the containing
  :t:`module` into :t:`scope`.

* :dp:`fls_ce73bg0BqV1X`
  Otherwise bring all :t:`entities <entity>` that the :t:`simple import path`
  resolves to that are visible from the location of the
  :t:`simple import` into :t:`scope`.
```

This is the semantic counterpart to the path-position rule in the separate `self` path thread. It keeps `use a::b::{self};`, `use a::{b::self};`, and `use ::std::{vec::self as std_vec};` valid, while direct final-`self` paths such as `use a::b::self as name;` remain rejected by the path-position rule.

I would also update `fls_iQOgxNihUEr7` so the underscore-import rule matches this definition:

```rst
:dp:`fls_iQOgxNihUEr7`
An :t:`entity` imported by a :t:`simple import` subject to a :t:`renaming` with
character underscore ``_`` is added into :t:`scope` without a :t:`name`.
```

The changelog should list `fls_WAA4WmohGu6T`, `fls_2bkcn83smy2y`, `fls_wRmvtgQkFA6w`, `fls_kz2Gij5wHXnl`, `fls_ar03D5rxjzy0`, and `fls_iQOgxNihUEr7` as changed; `fls_yY58pFpkig9o` as removed; and the :t:`simple import` glossary entry as updated. No extra paragraph ID should be needed if the `::` prefix preservation wording stays folded into `fls_WAA4WmohGu6T`.

src/entities-and-resolution.rst
PLeVasseur, line 1: Target: `src/entities-and-resolution.rst`, Paths legality rules near `fls_opn5n5t2mo3m`.

I think this general path-position rule should also get the Rust 2021 use-path exception for braced/nesting-import final-`self` imports.

The ordinary path rule remains that `self` is only used at the beginning of a path. Use declarations have an additional `self` import behavior inside brace syntax: a final `self` in a nesting import imports the parent entity selected by the fully constructed import prefix. rust-lang/rust#146972 applies this use-path behavior to Rust 2018 and later, which covers the FLS Rust 2021 edition scope.

That keeps braced prefixed imports valid, including `use crate::foo::bar::{self};`, `use crate::foo::bar::{self as name};`, `use crate::foo::{bar::foobar::quxbaz::self};`, and `use crate::foo::{bar::foobar::quxbaz::self as name};`. It still rejects direct final-`self` paths such as `use crate::foo::bar::self as name;`, initial-`self` direct final paths such as `use self::self as name;`, and intermediate-`self` paths such as `use foo::self::bar;` and `use a::{b::self::c};`.

Replace `fls_opn5n5t2mo3m` with:

```rst
:dp:`fls_opn5n5t2mo3m`
If a :t:`path segment` is expressed as either :t:`keyword` ``crate``,
:t:`keyword` ``$crate``, or :t:`keyword` ``Self``, then the
:t:`path segment` shall be the first :t:`path segment` of a :t:`path`.
If a :t:`path segment` is expressed as :t:`keyword` ``self``, then the
:t:`path segment` shall be the first :t:`path segment` of a :t:`path`, except
that it may be the last :t:`path segment` of the :t:`simple path` of a
:t:`simple import` that appears in a :t:`nesting import`.
```

This keeps `self::foo` and `use self as name;` valid, keeps `use a::b::{self as name};` and `use a::{b::self as name};` valid for the use-import parent-entity case, and still rejects `use a::b::self as name;`, `use self::self as name;`, `use a::{self::b};`, and `foo::self::bar`.

I checked the following with `rustc +1.95.0 --edition=2021`:

```rust
// Valid.
use crate::foo::bar::{self as bar_alias};
use crate::foo::{bar::foobar::quxbaz::self};
use crate::foo::{bar::foobar::quxbaz::self as quxbaz_alias};
use {self::self as this_module};
use ::std::{vec::self as std_vec};

// Invalid.
use crate::foo::bar::self as bar_alias;
use self::self as this_module;
use a::{self::b};
use a::{b::self::c as bad};
```

Since this changes a general path legality paragraph, list `fls_opn5n5t2mo3m` as changed in the Rust 1.95 keyword-import changelog entry. The corresponding simple-import semantics should still get the `fls_wRmvtgQkFA6w` update from the simple-import thread so final-`self` nesting imports bring the parent entity selected by the import prefix into scope.

PLeVasseur (changes requested):
I rechecked this against rust-lang/rust#146972, rust-lang/reference#2136, the current Reference text, and the FLS path/import model. The PR is pointed at the right upstream change. I think there are a few alignment points left where the new restriction paragraphs and the older path/import rules would otherwise disagree.

The main thing I am trying to preserve is that the new restriction paragraphs line up with the surrounding path/import rules. The current Reference text and the post-rust-lang/rust#146972 resolver model also rely on the surrounding `self`, `super`, `$crate`, and `::` path rules lining up with the import restrictions. The FLS currently still has older rules that conflict with the new paragraphs, especially around `use self as name;`, `use self::{self};`, braced parent-entity imports such as `use a::b::{self as name};` and `use a::{b::self as name};`, and `use self::super as name;`.

The remaining alignment points I see are:

1. Use the meeting outcome/Hristian follow-up wording: the :t:`entity` imported by the related :t:`simple import` is subject to the :t:`renaming`, not the textual path segment.
2. Update the existing `self` import rules so renamed current-module imports are accepted and unrenamed current-module imports remain rejected.
3. Update `fls_opn5n5t2mo3m` for Rust 2021 use paths so `self` is still restricted in ordinary paths, while braced/nesting-import parent-entity imports such as `use a::b::{self};`, `use a::b::{self as name};`, `use a::{b::self};`, and `use a::{b::self as name};` remain valid. Direct final-`self` paths such as `use a::b::self as name;` are still rejected by Rust 1.95.0 with E0429; rust-lang/rust#146972 leaves general support for `self` at the end of paths as future work.
4. Update `fls_7k88ypcgaoff` so `super` is permitted to follow an initial `self`, while still rejecting non-leading `super` in paths such as `foo::super` or `crate::super`.
5. Preserve the existing FLS distinction between `crate` and `$crate`; `$crate` resolves to the crate that declares the macro being expanded, not simply to the current crate in the same way as `crate`.
6. Keep the Rust 2021 `::` rule narrow: `use ::std::io;` remains valid, while importing the extern-prelude root itself is not valid. The constructed import-prefix model should preserve leading `::` rather than treating a bare `::` prefix as an empty local prefix.
7. Avoid saying that a :t:`simple import` brings a path into scope. Paths are syntax; imports bring entities/names into scope. Also update the simple-import semantics for final-`self` nesting imports, make empty non-global import prefixes resolve to the current module, preserve global `::` import prefixes, and update the existing underscore-renaming rule so it is not trait-specific.
8. The enum-variant-through-type-alias changelog classification is addressed by moving `fls_LV94x3HlpBWk` to FLS corrections; the final changelog needs to account for the paragraph and glossary changes after this source-shape pass.

The changelog should account for IDs relative to the merge base rather than earlier PR revisions. With the source shape from the thread comments, `fls_sUhnfV62HJrb`, `fls_QGdeRTe0H1Uc`, `fls_aam34hsRmKU2`, `fls_P6dFw89ZDKv2`, and `fls_61qKMG5tuv12` are new paragraphs; `fls_opn5n5t2mo3m`, `fls_WAA4WmohGu6T`, `fls_2bkcn83smy2y`, `fls_iuzvtr3oax1o`, `fls_90hQvSh7Bfyg`, `fls_wRmvtgQkFA6w`, `fls_kz2Gij5wHXnl`, `fls_ar03D5rxjzy0`, `fls_iQOgxNihUEr7`, `fls_RUiFQ17bmRLt`, and `fls_7k88ypcgaoff` are changed; `fls_cw006jhlboa`, `fls_yY58pFpkig9o`, `fls_Pxc0Ts8Y7pfW`, and `fls_hv3xT2CjZuxc` are removed; and the :t:`simple import` glossary entry is updated.

I trial-applied this source shape together, expanded the Rust 1.95.0 edition-2021 test matrix for braced final-`self` imports, and `./make.py --clear` builds. The snippets should be copy-paste-ready as a combined patch.

I left the detailed suggestions in individual threads/comments; they are intended to be applied as one source shape.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..29ca7f81ca4d9b9ff38d8a2e02309f1daa553835)*

src/entities-and-resolution.rst
tshepang: "the entity imported by the related simple import" is more wordy without adding precision, compared to "the imported entity". The word *related* also causes a pause in my mind, making we wonder what is related, that I could have missed something. It makes it a bit more confusing, making me adjust to the fact that it describes the import that is in the same sentence.

I do not agree that we need to use separate wording for `crate` and `$crate`, because the fact that the latter can only be used in macros should already be apparent elsewhere (`fls_774uryecc2sx`). I already consider "to import the current crate" as overspecification (and therefore confusing), because you can't do `use crate as foo` for anything but importing the current crate.

I made d6f50e9 to follow my thinking.

src/entities-and-resolution.rst
tshepang: see https://github.com/rust-lang/fls/pull/694#discussion_r3407592484

src/entities-and-resolution.rst
kirtchev-adacore: I actually like these suggestions, however `fls_opn5n5t2mo3m` should be split into two:
```
:dp:`fls_opn5n5t2mo3m`
If a :t:`path segment` is expressed as either :t:`keyword` ``crate``,
:t:`keyword` ``$crate``, or :t:`keyword` ``Self``, then the
:t:`path segment` shall be the first :t:`path segment` of a :t:`path`.

<new_FLS_ID>
If a :t:`path segment` is expressed as :t:`keyword` ``self``, then the
:t:`path segment` shall be the first :t:`path segment` of a :t:`path`, except
that it may be the last :t:`path segment` of the :t:`simple path` of a
:t:`simple import` that appears in a :t:`nesting import`.
```

tshepang: adjusted the wording for new rule slightly 9e87caf9562f00d766a3b562b43fe5637d7a6983

src/entities-and-resolution.rst
tshepang: I have simplified this to a2fd6c1

src/entities-and-resolution.rst
tshepang: 8bc4ab2

src/entities-and-resolution.rst
tshepang: 001ba7440c553790951e6c9ece0c4ad82a503791 expands `fls_aam34hsRmKU2` suggestion, while removing "import path prefix" term since I don't quite understand the definition.

I also added `fls_aam34hsRmKU2` as suggested, though I do not understand it well, particularly "candidate external prelude entity". That is, I don't understand the explanations given... maybe they need a rephrase.

kirtchev-adacore: https://rust-lang.github.io/fls/entities-and-resolution.html#term_candidate_external_prelude_entity

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1117,21 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
 If two :t:`[glob import]s` import the same :t:`entity` under the same :t:`name`,
 the :t:`visibility` of the :t:`name` is the most permissive one.
 
+:dp:`fls_RUiFQ17bmRLt`
+A :t:`simple import` with a single :t:`path segment` of :t:`keyword` ``self`` shall be subject to a :t:`renaming`.
```
kirtchev-adacore:
```suggestion
A :t:`simple import` with a single :t:`path segment` expressed as :t:`keyword` ``self`` shall be subject to a :t:`renaming`.
```
The "expressed as" form is used throughout this section, so it is good to be consistent.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..001ba7440c553790951e6c9ece0c4ad82a503791)*

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1117,21 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
+
+:dp:`fls_sUhnfV62HJrb`
+An :t:`import <use import>` with a single :t:`path segment` of either :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` shall be subject to a :t:`renaming`.
```
kirtchev-adacore:
```suggestion
A :t:`use import` with a single :t:`path segment` expressed as either :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` shall be subject to a :t:`renaming`.
```
Since the FLS is supposed to be more precise, `use import` is preferable over just `import`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..001ba7440c553790951e6c9ece0c4ad82a503791)*

src/entities-and-resolution.rst
tshepang: 56129b0f79c55724520d3c9476e3c6ebbad97a3e is for "simple import"... I removed the renaming thing since it already is covered by `fls_FILuR3pfwjw3` and `fls_iQOgxNihUEr7`

tshepang: - does this text refer to a `use {segment}`: `An import path prefix with no path segments and no namespace qualifier resolves to the current module`

- in `An import path prefix that starts with namespace qualifier :: preserves that namespace qualifier in the constructed prefix`, what does "preserve" mean

as a bonus question, what is a "constructed prefix"

tshepang: e3a53f6e70b585a1105f268c60c4b16f3cb19c84 takes the suggestion that tightens terminology, where we now use "path segment" term more (in this pr), and flattens the list (to 1 level deep instead of 2 levels)

tshepang: 55ab2bc4259aa87bbe1a6efa203b8fcd29eeee27 for the `_`

src/entities-and-resolution.rst
```diff
@@ -1049,9 +1044,7 @@ An :dt:`import path prefix` is the fully constructed :t:`path` prefix of a
    the current :t:`use import`.
 
 :dp:`fls_2bkcn83smy2y`
-A :t:`simple import` is a :t:`use import` that brings all :t:`entities <entity>`
-it refers to into scope, optionally with a different
-:t:`name` than they are declared with by using a :t:`renaming`.
+A :dt:`simple import` is a :t:`use import` that bring an :t:`entity` selected by its :t:`simple import path`, or by its :t:`import path prefix` when its :t:`simple path` ends in :t:`keyword` ``self`` and it appears in a :t:`nesting import`, into :t:`scope`.
```
kirtchev-adacore:
```suggestion
A :dt:`simple import` is a :t:`use import` that bring into :t:`scope` an :t:`entity` selected by its :t:`simple import path`, or by its :t:`import path prefix` when its :t:`simple path` ends in :t:`keyword` ``self`` and it appears in a :t:`nesting import`.
```
In addition, there are too many candidates for the `it` in "and **it** appears in a testing import". The `it` could be:
- Keyword `self`
- The entity
- The import path prefix
- The simple import
- The simple import path
- The simple path

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..55ab2bc4259aa87bbe1a6efa203b8fcd29eeee27)*

src/entities-and-resolution.rst
kirtchev-adacore: I second Tshepang's questions [above](https://github.com/rust-lang/fls/pull/694#discussion_r3468781913).

src/entities-and-resolution.rst
```diff
@@ -1075,37 +1068,30 @@ A :t:`glob import` brings :t:`[name]s` into :t:`scope` as follows:
...
 
-* :dp:`fls_kz2Gij5wHXnl`
-  If the :t:`simple path` is keyword ``self`` and:
-
-  * :dp:`fls_yY58pFpkig9o`
-    The :t:`simple import` is in a :t:`nesting import`, then bring the
-    :t:`entity` in :t:`type namespace` that the :t:`import path prefix` resolves
-    to into :t:`scope`.
+* :dp:`fls_yY58pFpkig9o`
+  If the :t:`simple import` appears in a :t:`nesting import`, and the last :t:`path segment` of its :t:`simple path` is expressed as :t:`keyword` ``self``, then bring the :t:`entity` in :t:`type namespace` that the :t:`import path prefix` resolves to into :t:`scope`.
```
kirtchev-adacore:
```suggestion
  If the :t:`simple import` appears in a :t:`nesting import`, and the last :t:`path segment` of its :t:`simple path` is expressed as :t:`keyword` ``self``, then the :t:`simple import` brings the :t:`entity` that the :t:`import path prefix` resolves to in  :t:`scope` of the :t:`type namespace`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..55ab2bc4259aa87bbe1a6efa203b8fcd29eeee27)*

src/entities-and-resolution.rst
```diff
@@ -1075,37 +1068,30 @@ A :t:`glob import` brings :t:`[name]s` into :t:`scope` as follows:
...
 
-  * :dp:`fls_ar03D5rxjzy0`
-    Otherwise bring the containing :t:`module` into :t:`scope`.
+* :dp:`fls_ar03D5rxjzy0`
+  If the :t:`simple path` is :t:`keyword` ``self``, then bring the containing :t:`module` into :t:`scope`.
```
kirtchev-adacore:
```suggestion
  If the :t:`simple path` is expressed as :t:`keyword` ``self``, then the :t:`simple import` brings the containing :t:`module` into :t:`scope`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..55ab2bc4259aa87bbe1a6efa203b8fcd29eeee27)*

src/entities-and-resolution.rst
```diff
@@ -295,24 +295,25 @@ A :t:`path segment` is an element of a :t:`path`.
 A :t:`path` is subject to :t:`path resolution`.
 
 :dp:`fls_opn5n5t2mo3m`
-If a :t:`path segment` is expressed as either :t:`keyword` ``crate``,
-:t:`keyword` ``$crate``, :t:`keyword` ``self``, or :t:`keyword` ``Self``, then
-the :t:`path segment` shall be the first :t:`path segment` of a :t:`path`.
+If a :t:`path segment` is expressed as either :t:`keyword` ``crate``, :t:`keyword` ``$crate``, or :t:`keyword` ``Self``, then the :t:`path segment` shall be the first :t:`path segment` of a :t:`path`.
+
+:dp:`fls_uSajfdSsbxna`
+If a :t:`path segment` is expressed as :t:`keyword` ``self``, then the :t:`path segment` shall be the first :t:`path segment` of a :t:`path`, except that it may also be the last :t:`path segment` of a :t:`simple import` that appears in a :t:`nesting import`.
```
kirtchev-adacore:
```suggestion
If a :t:`path segment` is expressed as :t:`keyword` ``self``, then the :t:`path segment` shall be either the first :t:`path segment` of a :t:`path`, or the last :t:`path segment` of a :t:`simple import` that appears in a :t:`nesting import`.
```
I think this reads a bit better.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..55ab2bc4259aa87bbe1a6efa203b8fcd29eeee27)*

src/entities-and-resolution.rst
tshepang: e437648

src/entities-and-resolution.rst
PLeVasseur: Yes, the idea is that the sentence is meant to cover imports with no prefix before the braces. For example:

```rust
use {segment as segment_alias};
use {self as this_module};
```

In those cases, the import path prefix has no path segments and no leading `::`, so it is a local empty prefix and resolves to the current module.

That is the same prefix shape as `use {segment};`; the alias just avoids the unrelated duplicate-name error you can get when re-importing a module under its existing name.

The contrast is:

```rust
use ::{self as root};
```

That also has no path segments before `self`, but it does have the `::` namespace qualifier. So it is not the empty local prefix; it is the global extern-prelude root, which Rust rejects importing directly.

On terminology: `import path prefix` is already a defined FLS term here. `constructed prefix` was just shorthand on my side for the result of constructing the `import path prefix`; we can avoid that phrase in final wording.

By "preserve", I meant: do not drop the leading `::` while constructing the import path prefix.

This matters because `::` is a namespace qualifier, not a path segment. If the construction only carries over path segments, `::std` can accidentally become indistinguishable from local `std` in the FLS model.

Clearer wording than "preserves" would be:

```rst
If the :t:`simple path prefix` used to construct an :t:`import path prefix`
starts with :t:`namespace qualifier` ``::``, then the resulting
:t:`import path prefix` starts with :t:`namespace qualifier` ``::``.
```

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1110,21 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
 If two :t:`[glob import]s` import the same :t:`entity` under the same :t:`name`,
 the :t:`visibility` of the :t:`name` is the most permissive one.
 
+:dp:`fls_RUiFQ17bmRLt`
+A :t:`simple import` with a single :t:`path segment` expressed as :t:`keyword` ``self`` shall be subject to a :t:`renaming`.
```
PLeVasseur: This seems pretty close, but I think it still misses one current-module import spelling.

The rule now catches single-segment forms like:

```rust
use self;
use {self};
```

But rustc also rejects this unrenamed current-module import:

```rust
use {self::self};
```

That path has two `self` segments, so it is not covered by "a single path segment expressed as `self`". The restriction should likely be made about importing the current module without a renaming, not only about a single textual segment.

Suggested shape:

```rst
When a :t:`path segment` expressed as :t:`keyword` ``self`` is used to import
the current :t:`module`, the imported :t:`entity` shall be subject to a
:t:`renaming`.
```

This still allows parent-entity imports like `use a::b::{self};`, because those import `a::b`, not the current module.

src/entities-and-resolution.rst
```diff
@@ -1135,6 +1110,21 @@ same :t:`namespace` but refer to different :t:`entities <entity>` if the
...
+
+:dp:`fls_sUhnfV62HJrb`
+A :t:`use import` with a single :t:`path segment` expressed as either :t:`keyword` ``crate`` or :t:`keyword` ``$crate`` shall be subject to a :t:`renaming`.
+
+:dp:`fls_QGdeRTe0H1Uc`
+When a :t:`path segment` expressed as :t:`keyword` ``super`` is used to import a parent :t:`module`, the imported :t:`entity` shall be subject to a :t:`renaming`.
+
+:dp:`fls_aam34hsRmKU2`
+A :t:`global path` where any of its :t:`[path segment]s` are expressed as any of the :t:`[keyword]s`, ``self``, ``super``, ``crate``, and ``$crate``, shall not be used.
```
PLeVasseur: I think this rule might be a bit broad now.

It should reject importing the bare global/extern-prelude root:

```rust
use ::{self as root}; // invalid
```

But it should not reject every global path that later contains `self`. This remains valid because `std` is the extern-prelude crate and the final `self` is the nesting-import shorthand for `std::vec`:

```rust
use ::std::{vec::self as std_vec}; // valid
```

`fls_P6dFw89ZDKv2` already covers the general rule that a Rust 2021 global path starts with an extern-prelude crate name. I think this paragraph should stay narrow and cover only the bare-root import case, for example:

```rst
A :t:`simple import` whose :t:`import path prefix` consists only of
:t:`namespace qualifier` ``::`` and whose :t:`simple path` consists of a single
:t:`path segment` expressed as :t:`keyword` ``self`` shall not be used.
```

By the way, I am not trying to drop the `crate` / `super` / `$crate` cases. Those should fall out of the general global-path rule (`fls_P6dFw89ZDKv2`): after leading `::`, the first path segment must be an extern-prelude crate identifier, so `::crate`, `::super`, `::self`, and `::$crate` are already rejected. This paragraph seems to only need to cover the separate final-`self` nesting-import case where the prefix is bare `::`, as in `use ::{self as root};`.

src/entities-and-resolution.rst
```diff
@@ -1075,37 +1068,30 @@ A :t:`glob import` brings :t:`[name]s` into :t:`scope` as follows:
...
 
-* :dp:`fls_kz2Gij5wHXnl`
-  If the :t:`simple path` is keyword ``self`` and:
-
-  * :dp:`fls_yY58pFpkig9o`
-    The :t:`simple import` is in a :t:`nesting import`, then bring the
-    :t:`entity` in :t:`type namespace` that the :t:`import path prefix` resolves
-    to into :t:`scope`.
+* :dp:`fls_yY58pFpkig9o`
+  If the :t:`simple import` appears in a :t:`nesting import`, and the last :t:`path segment` of its :t:`simple path` is expressed as :t:`keyword` ``self``, then the :t:`simple import` brings the :t:`entity` that the :t:`import path prefix` resolves to in :t:`scope` of the :t:`type namespace`.
```
PLeVasseur: The intended semantics are right, but this seems a bit hard to parse for me when it says the entity is brought "in scope of the type namespace".

I think it might read a little cleaner if we said that the prefix resolves in the type namespace, and the resulting entity is brought into scope. Something like the below, maybe.

```rst
If the :t:`simple import` appears in a :t:`nesting import` and the last
:t:`path segment` of its :t:`simple path` is expressed as :t:`keyword`
``self``, then the :t:`simple import` brings the :t:`entity` in
:t:`type namespace` that the :t:`import path prefix` resolves to into
:t:`scope`.
```

Example:

```rust
use a::b::{self as b_alias};
```

Here final `self` imports the entity selected by prefix `a::b`, not a child item named `self`.

PLeVasseur (changes requested):
Thanks, this has handled most of my earlier review. I think the remaining issues are narrower now, so I tried to leave some focused comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/4bfb88c246a56e495bd49218554a77eeb4493e47..f6cf2fce5c0816a6e585f8ca552e14c7c2915921)*

src/entities-and-resolution.rst
tshepang: I don't see the braces in `use {segment}` as semantically meaningful, and that makes me hesitate to consider that import as having an empty import path prefix. Similarly, the braces of `use {{{segment}}}` are also not semantically meaningful (else we would have to consider that import as having 3 import path prefixes).

As an aside, in `use a::{b, c::d}`, would you consider `a` to be the only path prefix, or would you consider that to include `a.c`? What about `use a.b.c`... is the path prefix `a` or `a.b`?

src/entities-and-resolution.rst
tshepang: 57ceb2fa962ce348555a927ea4afac464306c2e7

src/entities-and-resolution.rst
tshepang: 70947a6edbab34462ba0286bc2cd7f25adceddf2

src/entities-and-resolution.rst
tshepang: 4332f41

src/entities-and-resolution.rst
PLeVasseur: We discussed the second paragraph in the meeting and agreed that according to the current definition of [path prefix](https://rust-lang.github.io/fls/entities-and-resolution.html#:~:text=14%2E3%3A9,stripped) in the FLS, for these examples:
> As an aside, in `use a::{b, c::d}`, would you consider `a` to be the only path prefix, or would you consider that to include `a.c`? What about `use a.b.c`... is the path prefix `a` or `a.b`?

We'd end up with:
- `b`'s path prefix is `a`
- `d`'s path prefix is `a::c`

tshepang: There is also confusion (for me) what "prefix" means, for we have these differing definitions:

- "A simple path prefix is the leading simple path of a glob import or a nesting import", meaning path prefix of `use a::b::{c, d::{e, f}}` is `a::b`
- "A path prefix is a path with its last path segment and namespace qualifier stripped", [shown above](https://github.com/rust-lang/fls/pull/694#discussion_r3499679044)

The former should maybe be renamed to "common path prefix", which would more closely match the term used in Reference, "common prefix"

PLeVasseur: Regarding:
> I don't see the braces in `use {segment}` as semantically meaningful, and that makes me hesitate to consider that import as having an empty import path prefix. Similarly, the braces of `use {{{segment}}}` are also not semantically meaningful (else we would have to consider that import as having 3 import path prefixes).

I'm a bit worried that the distinction I'm attempting to draw here might not be landing correctly.

The braces are only grouping syntax; the semantic distinction is the prefix before the brace group. In Rust 2021, no prefix resolves from the current scope/module, bare `::` attempts to import the extern-prelude root and is rejected, and `::std` is a valid global path through the extern prelude.

| Identifier | Rust 2021 example | Result | Why |
| --- | --- | --- | --- |
| `A` | `use {self as this_module};` | Compiles | No prefix before braces; imports the current module with required renaming. |
| `B` | `use {self};` | Fails | Same no-prefix/current-module case, but missing required renaming. |
| `C` | `use ::{self as root};` | Fails | Bare `::` before braces tries to import the extern-prelude root. |
| `D` | `use ::std::{self as std_alias};` | Compiles | `::std` is a real global path through the extern prelude; it is not the bare root. |

Taking them 1-by-1:

#### `A`

Take this example ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=92d11dc9e06b6997b67d70735bdc1de8)):
```rust
use {self as this_module};

fn main() {
    let _ = this_module::main as fn();
}
```

This compiles under Rust 2021.

Imports the current module with required renaming.

#### `B`

Take this example ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a4e91e123e982087b369f66cc6087131)):
```rust
// Fails with --edition=2021: error: extern prelude cannot be imported
use ::{self};

fn main() {}
```

This fails to compile under Rust 2021 with:
```
error: imports need to be explicitly named
 --> src/main.rs:2:6
  |
2 | use {self};
  |      ^^^^
  |
help: try renaming it with a name
  |
2 | use {self as name};
  |           +++++++
```

#### `C`

Take this example ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=920a72cbb0e44da069e807c4d942e14e)):
```rust
// Fails with --edition=2021: extern prelude cannot be imported
use ::{self as root};

fn main() {}
```

This fails to compile under Rust 2021 with: extern prelude cannot be imported

#### `D`

Take this example ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8b5db9f50a1f577dbc08b4f2e9b77f8c)):
```rust
// Compiles with --edition=2021
use ::std::{self as std_alias};

fn main() {
    let _: std_alias::vec::Vec<u8> = std_alias::vec::Vec::new();
}
```

This compiles under Rust 2021.

`::std` is a real global path through the extern prelude; it is not the bare root.

tshepang: as an aside, note that for D, this is also valid syntax, `use {{::{{std::{self as std_alias}}}}};`, which (to me) indicates that the brace has 2 roles: mere visual grouping and semantic grouping

tshepang: other than following [this proposal](https://github.com/rust-lang/fls/pull/694#discussion_r3499774990), am not sure how to move forward, short of re-writing a number of things in the chapter... I find much of the text hard to understand

tshepang: 79aefcb

src/entities-and-resolution.rst
```diff
@@ -1018,7 +1013,7 @@ A :t:`use import` brings :t:`entities <entity>` :t:`in scope` within the
 :t:`use import` resides.
 
 :dp:`fls_sxo1jb25pl8a`
-A :t:`simple path prefix` is the leading :t:`simple path` of a :t:`glob import`
+A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
 or a :t:`nesting import`.
 
 :dp:`fls_WAA4WmohGu6T`
```
PLeVasseur, line 1019: I think we can separate this from the remaining PR 694 work.

There is a real FLS issue here: the import-prefix model mostly describes prefixes in terms of path segments, but a leading ``::`` is not a path segment. That means the current text does not clearly preserve the difference between an empty local prefix, as in ``use {self as this_module};``, and a global/external-prelude prefix, as in ``use ::{self as root};`` or ``use ::std::{self as std_alias};``. This matters because dropping ``::`` can change resolution from extern-prelude lookup to local lookup, and it also makes rules like "the import path prefix consists only of ``::``" hard to apply rigorously.

After digging through the history, though, this looks like an existing FLS modeling problem rather than something introduced by this PR. The same issue is already present in the pre-PR ``SimplePathPrefix ::= SimplePath? "::"`` grammar and the existing ``fls_WAA4WmohGu6T`` construction rule, which builds prefixes from path segments. PR 694 mostly exposes the problem because it adds rules that refer directly to a bare ``::`` import prefix.

So I do not think we need to make this PR solve the whole import-prefix model.

Based on our discussion in last week's t-fls meeting we could do the following:

1. Keep this PR focused on the Rust 1.95 keyword-import restrictions.
2. Open a follow-up issue for the inherited import-prefix/leading-``::`` modeling problem.
3. Merge this PR

What do y'all think? @tshepang @kirtchev-adacore

PLeVasseur (approved):
Thinking over what we talked about last week, I could see the rationale of approving and merging this, and opening up a follow-up issue.

So -- I'm approving with the intention that is what we do.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/694/3c73fb2263ce3ffc9f7d7592a910d42372cad001..79aefcbd298729a14cdd5a152156f02e59580ae8)*

src/entities-and-resolution.rst
kirtchev-adacore: I like the approach.

PLeVasseur: Filed #711 to track the inherited leading-`::` import-prefix modeling issue. That leaves #694 scoped to the Rust 1.95 keyword-import restrictions as agreed.

kirtchev-adacore (approved):

PR #698, opened 2026-04-20
document Closure capture precision

tshepang:
This captures some gaps FLS has compared to Reference, and the content is adjusted from [Closure capture precision](https://doc.rust-lang.org/nightly/reference/types/closure.html#capture-precision) section.


src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+Move closures can only capture the prefix of a :t:`capture path` that runs up to, but not including, the first :t:`dereference` of a :t:`reference`.
+
+:dp:`fls_7NEEJgKSpQQ8`
+Closures will only capture the prefix of a :t:`capture path` that runs up to, but not including, the first :t:`dereference` of a :t:`raw pointer`.
+
+:dp:`fls_kYFd3p06pWWV`
+Closures will only capture the prefix of a :t:`capture path` of a :t:`union` that runs up to union itself.
+
+:dp:`fls_fATMTNUOHsfb`
+Closures will only capture the prefix of the :t:`capture path` that runs up to, but not including, the first :t:`field access expression` into a structure that uses the :t:`attribute` ``packed`` representation, in unalidged :t:`[field]s` in a struct.
```
Hoverbear:
```suggestion
Closures will only capture the prefix of the :t:`capture path` that runs up to, but not including, the first :t:`field access expression` into a structure that uses the :t:`attribute` ``packed`` representation, in unaligned :t:`[field]s` in a struct.
```
Typo?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/f5c48f51906cc12b5745ced6658d96ea0850552a..c732101f15b3c8265486e226bed3f31b6d035cbd)*

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
 :t:`trait` if all the :t:`[type]s` of the :t:`[value]s` of the :t:`capturing
 environment` implement the :std:`core::marker::Send` :t:`trait`.
 
+.. _fls_G64vdcIyB2Is:
+
+Closure capture precision
```
kirtchev-adacore: You may want to rename this section "Capture precision", similar to the Reference, and make it subheading of [6.22. Capturing](https://rust-lang.github.io/fls/expressions.html#capturing).

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:dp:`fls_rdDT7jsaOMbs`
```
kirtchev-adacore: This paragraph should be swapped with :dp:`fls_j9WyKVyOLFon`. The FLS avoids forward referencing as much as possible. Here you are using the term "place projection", but you have not defined it yet.

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+A :dt:`capture path` is a sequence starting with a :t:`variable` from the environment followed by zero or more :t:`[place projection]s` from that variable.
```
kirtchev-adacore:
```suggestion
A :dt:`capture path` is a sequence starting with a :t:`variable` from the :t:`capturing environment` followed by zero or more :t:`[place projection]s` from that :t:`variable`.
```

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+
+:dp:`fls_j9WyKVyOLFon`
+A :dt:`place projection` is a :t:`field access expression`, tuple :t:`field access expression`, :t:`dereference`, array or slice :t:`index expression`, or :t:`pattern` destructuring applied to a variable.
```
kirtchev-adacore:
```suggestion
A :dt:`place projection` is a :t:`field access expression`, :t:`dereference`, :t:`array` or :t:`slice` :t:`index expression`, or :t:`pattern` destructuring applied to a :t:`variable`.
```
I think "tuple :t:`field access expression`" is redundant because a "regular" field access expression subsumes tuple indexing.

Note that we lack a section on pattern destructuring in [Chapter 5. Patterns](https://rust-lang.github.io/fls/patterns.html#patterns).

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+
+:dp:`fls_TbfUxVf8PKPs`
+The closure :t:`[borrow]s` or :t:`moves <by move>` the :t:`capture path`, which may be truncated based on these rules:
```
kirtchev-adacore: How about

A :t:`closure expression` :t:`[borrow]s` or :t:`moves <by move>` the :t:`capture path`, as follows:

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+
+- :dp:`fls_4TESOxGpEY2h`
+  When a :t:`capture path` and one of the ancestors of that path are both captured by a closure, the ancestor path is captured with the highest capture mode among the two captures, using the strict weak ordering: :t:`immutable borrow` < :t:`unique immutable borrow` < :t:`mutable borrow` < :t:`move <by move>`.
```
kirtchev-adacore: It is unclear what an "ancestor" is. Be careful when using the term "path", because we also have these [paths](https://rust-lang.github.io/fls/entities-and-resolution.html#paths). What the Reference describes is already documented in [6.22:12 The capture mode ...](https://rust-lang.github.io/fls/expressions.html#capturing).

I would rephrase the paragraph as follows (note that it is part of a list!):

- When a :t:`capture path` and an ancestor :t:`capture path` and both :t:`captured`, the ancestor :t:`capture path` is :t:`captured` with the highest :t:`capture mode` among the two :t:`[capture path]s`.

Note that I intentionally dropped "by a closure" because a "capturing expression" is either an async block expression or a closure expression, and I suspect that these rules do apply when the context is an async block expression.

I also dropped the ordering because it is already documented by "capture mode". On a side note, perhaps 6.22:14 should be changed to "1. By immutable reference capture. (lowest precedence)" and 6:22:17 to "4. By value capture. (highest precedence)", or just use a small table similar to [expression precedence](https://rust-lang.github.io/fls/expressions.html#expression-precedence).

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+
+- :dp:`fls_eNkZWskzznW6`
+  The :t:`capture path` is truncated at the rightmost :t:`dereference` in the capture path if the dereference is applied to a :t:`shared reference`.
```
kirtchev-adacore:
```suggestion
- The :t:`capture path` is truncated at the rightmost :t:`dereference` in the :t:`capture path` if the :t:`dereference` is applied to a :t:`shared reference`.
```

src/types-and-traits.rst
```diff
@@ -903,6 +903,74 @@ A :t:`closure type` implicitly implements the :std:`core::marker::Sync`
...
+
+:dp:`fls_v8IFXHJnXhez`
+A :t:`place` is not captured when an :t:`underscore expression` is used to bind it.
```
kirtchev-adacore: The Reference is unclear what "place" refers to - is it a "place projection", a "place expression", or a "place". I suspect the former, but this needs confirmation. Once that is cleared up, there are options to refactor some of the rules below.

kirtchev-adacore (changes requested):
This is it for now. I will have to revisit the remaining rules once the "place" business has been cleared up. I think we can find better wording/grouping of the remaining rules, as they feel somewhat randomly sprinkled.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/f5c48f51906cc12b5745ced6658d96ea0850552a..c0edef222591c64c0d1f484622e4d0c7565f4659)*

src/types-and-traits.rst
tshepang: heh, I missed that 6.22 even existed :hide:

src/types-and-traits.rst
tshepang: 40847f44d558b5bd022e0dd4d0729a96941da1b9

tshepang:
all addressed but https://github.com/rust-lang/fls/pull/698#discussion_r3115942684

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+~~~~~~~~~~~~~~~~~
+
+:dp:`fls_j9WyKVyOLFon`
+A :dt:`place projection` is a :t:`field access expression`, :t:`dereference`, :t:`array` or :t:`slice` :t:`index expression`, or :t:`pattern` destructuring applied to a :t:`variable`.
+
+:dp:`fls_rdDT7jsaOMbs`
+A :dt:`capture path` is a sequence starting with a :t:`variable` from the :t:`capturing environment` followed by zero or more :t:`[place projection]s` from that :t:`variable`.
+
+:dp:`fls_TbfUxVf8PKPs`
+A :t:`closure expression` :t:`[borrow]s` or :t:`moves <by move>` the :t:`capture path`, which may be truncated based on these rules:
```
kirtchev-adacore:
```suggestion
A :t:`closure expression` :t:`[borrow]s` or :t:`moves <by move>` the :t:`capture path`, as follows:
```
The FLS avoids using the term "rules", as everything pretty much is a rule.

Note that there is no need to mention truncation here because there is a dedicated rule further below.

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+
+- :dp:`fls_4TESOxGpEY2h`
+  When a :t:`capture path` and an ancestor :t:`capture path` are both :t:`captured <capturing>`, the ancestor :t:`capture path` is captured with the highest :t:`capture mode` among the two :t:`[capture path]s`.
```
kirtchev-adacore:
```suggestion
- When a :t:`capture path` and an ancestor :t:`capture path` are both :t:`captured <capturing>`, the ancestor :t:`capture path` is :t:`captured` with the highest :t:`capture mode` among the two :t:`[capture path]s`.
```
This and the following rules should be list items.

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+
+- :dp:`fls_eNkZWskzznW6`
+  The :t:`capture path` is truncated at the rightmost :t:`dereference` in the :t:`capture path` if the :t:`dereference` is applied to a :t:`shared reference`.
```
kirtchev-adacore, line 5369:
```suggestion
- The :t:`capture path` is truncated at the rightmost :t:`dereference` in the :t:`capture path` if the :t:`dereference` is applied to a :t:`shared reference`.
```
List item.

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+Closures will only capture the prefix of a :t:`capture path` that runs up to, but not including, the first :t:`dereference` of a :t:`raw pointer`.
+
+:dp:`fls_kYFd3p06pWWV`
+Closures will only capture the prefix of a :t:`capture path` of a :t:`union` that runs up to the union itself.
+
+:dp:`fls_fATMTNUOHsfb`
+Closures will only capture the prefix of the :t:`capture path` that runs up to, but not including, the first :t:`field access expression` into a structure that uses the :t:`attribute` ``packed`` representation, in unaligned :t:`[field]s` in a struct.
+
+:dp:`fls_fITor3jpmgrl`
+Taking the address of an unaligned :t:`field` captures the entire struct.
```
kirtchev-adacore: I made significant reorderings and rewording of some of the paragraphs.
```suggestion
:dp:`fls_Vt9C9mKxHOwo`
A :t:`place` is :t:`captured` by :t:`immutable borrow` if its :t:`discriminant` is read by :t:`pattern matching`.

:dp:`fls_v8IFXHJnXhez`
A :t:`place` is not :t:`captured` when an :t:`underscore expression` is used to bind it.

:dp:`fls_gujpU7p5n9Zx`
A :t:`place` is not :t:`captured` by destructuring :t:`[struct]s`, :t:`[tuple]s`, and :t:`[enum]s` with a single :t:`enum variant`.

:dp:`fls_t8tFLUg8O83Q`
A :t:`place` is not :t:`captured` by being matched against a :t:`rest pattern`.

:dp:`fls_RaONmCLH2KGM`
An :t:`array` or :t:`slice` is :t:`captured` whole.

:dp:`fls_Fs12dmznjsMf`
Matching against an :t:`enum variant` of an :t:`enum` with more than one :t:`[enum variant]s` :t:`captures` the :t:`place` by :t:`immutable borrow`.

:dp:`fls_7EXHdE2eOVek`
Matching against an :t:`enum variant` of an :t:`enum` with one :t:`enum variant` does not :t:`capture` the :t:`place`, unless it is subject to :t:`attribute` ``non_exhaustive``, in which case the :t:`place` is captured by :t:`immutable borrow`.

:dp:`fls_iLH8X2U4ADHb`
Matching against a :t:`range pattern` :t:`captures` the :t:`place` by :t:`immutable borrow`.

:dp:`fls_HMJUXHrvOmPl`
Matching a :t:`slice` against a :t:`slice pattern`, other than one with only a single :t:`rest pattern`, :t:`captures` the :t:`slice` by :t:`immutable borrow`.

:dp:`fls_Gj1znNpthHY6`
Matching an :t:`array` against a :t:`slice pattern` does not :t:`capture` the :t:`place`.

:dp:`fls_IFyJvb6mlFU4`
A :t:`closure expression` subject to keyword ``move`` :t:`captures` the prefix of a :t:`capture path` that runs up to, but not including, the first :t:`dereference` of a :t:`reference`.

:dp:`fls_7NEEJgKSpQQ8`
A :t:`closure expression` :t:`captures` the prefix of a :t:`capture path` that runs up to, but not including, the first :t:`dereference` of a :t:`raw pointer`.

:dp:`fls_kYFd3p06pWWV`
A :t:`closure expression` :t:`captures` the prefix of a :t:`capture path` of a :t:`union` that runs up to the :t:`union` itself.

:dp:`fls_fATMTNUOHsfb`
A :t:`closure expression` :t:`captures` the prefix of the :t:`capture path` that runs up to, but not including, the first :t:`field access expression` into a :t:`struct` that uses the :t:`attribute` :c:`repr` with modifier ``packed``, in unaligned :t:`[field]s` in a :t:`struct`.

:dp:`fls_fITor3jpmgrl`
Taking the address of an unaligned :t:`field` captures the entire struct.
```

kirtchev-adacore (changes requested):
I will probably need to read [Capture precision](https://doc.rust-lang.org/nightly/reference/types/closure.html#capture-precision) to ensure that no rules have been omitted.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..06229e769561c51aa70defc2bd4a9b2854804ba9)*

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+Move closures can only capture the prefix of a :t:`capture path` that runs up to, but not including, the first :t:`dereference` of a :t:`reference`.
+
+:dp:`fls_7NEEJgKSpQQ8`
+Closures will only capture the prefix of a :t:`capture path` that runs up to, but not including, the first :t:`dereference` of a :t:`raw pointer`.
+
+:dp:`fls_kYFd3p06pWWV`
+Closures will only capture the prefix of a :t:`capture path` of a :t:`union` that runs up to the union itself.
+
+:dp:`fls_fATMTNUOHsfb`
+Closures will only capture the prefix of the :t:`capture path` that runs up to, but not including, the first :t:`field access expression` into a structure that uses the :t:`attribute` ``packed`` representation, in unaligned :t:`[field]s` in a struct.
```
PLeVasseur: This does not currently match the Reference. Upstream truncates at the first field access into any `repr(packed)` struct, and explicitly says this includes even fields that are currently aligned, for future-compatibility reasons.

The current wording reads as if the rule only applies to unaligned fields, so it narrows the capture rule.

Support:
- Reference `type.closure.capture.precision.unaligned`

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
 
+.. _fls_G64vdcIyB2Is:
+
+Capture precision
+~~~~~~~~~~~~~~~~~
+
+:dp:`fls_j9WyKVyOLFon`
+A :dt:`place projection` is a :t:`field access expression`, :t:`dereference`, :t:`array` or :t:`slice` :t:`index expression`, or :t:`pattern` destructuring applied to a :t:`variable`.
+
+:dp:`fls_rdDT7jsaOMbs`
```
PLeVasseur: Two semantic gaps still seem unresolved here.

- `place projection` still seems narrower than the Reference because it omits automatic dereferences. Upstream says `dereference (and automatic dereferences)`, and that matters because these rules operate on the projected place, not only explicit `*` syntax.
- This subsection should also say how `capture path` relates to the existing `capture target` / `capturing environment` model above. Earlier in 6.22, a capture target is only a variable or a field of a variable; here, capture paths now include dereferences, indexing, raw-pointer truncation, union truncation, and packed-struct truncation, so it is hard to tell whether `capture path` refines that model or replaces it for this subsection.

Could this subsection tighten both definitions together, perhaps by widening `place projection` and adding a short bridge sentence about how `capture path` fits the model above?

Support:
- Reference `type.closure.capture.precision.place-projection`
- Reference `type.closure.capture.precision`
- `src/expressions.rst:5103-5184`
- `src/glossary.rst:768-785`

src/changelog.rst
```diff
@@ -36,6 +36,34 @@ Language changes in Rust 1.95.0
 - `Const blocks are no longer evaluated to determine if expressions involving fallible operations can implicitly be constant-promoted <https://github.com/rust-lang/rust/pull/150557>`_
 - `Make operational semantics of pattern matching independent of crate and module <https://github.com/rust-lang/rust/pull/150681>`_
 
+  Changed paragraphs:
```
PLeVasseur: The paragraph-ID list itself looks fine, but this changelog attribution still seems too broad.

The changed capture-mode paragraphs fit rust-lang/rust#150681, but the new capture-precision block spans multiple sources: some items align with the later pattern/discriminant work in rust-lang/rust#138961 and rust-lang/reference#1837, while others are pre-existing Reference behavior now being documented here.

Could this entry be split, or reworded so `#150681` is not presented as introducing the whole block? The note that these paragraphs `mostly document behavior that existed before this release` already points in that direction.

Support:
- rust-lang/rust#138961
- rust-lang/reference#1837
- rust-lang/rust#150681

src/expressions.rst
PLeVasseur: Separate completeness point on this section: it still seems to need the Reference's `Box` vs other `Deref` exception.

Precise capture through `(*b).field` is special-cased for `Box`; other `Deref` types like `Rc` capture the dereference base instead.

Could this section add a short rule or note making that distinction explicit? Otherwise the current summary overstates capture precision through dereference projections.

Support:
- Reference `type.closure.capture.precision.box-deref`
- Reference `type.closure.capture.precision.box-non-move.not-moved`
- Reference `type.closure.capture.precision.box-non-move.moved`
- Reference `type.closure.capture.precision.box-move.read`

src/expressions.rst
PLeVasseur: The shared-prefix rule also seems to need the Reference's recursive-propagation note.

A deeper descendant can strengthen an intermediate ancestor, and that stronger mode may then need to propagate again to a more distant ancestor.

Could a short note be added here saying this rule may need to be applied recursively?

Support:
- Reference `type.closure.capture.precision.shared-prefix`

src/expressions.rst
PLeVasseur: Several rules in this block still seem too strong relative to the Reference.

- `_` here should use the FLS term `underscore pattern`, not `underscore expression`, and `_` does not introduce a binding.
- The destructuring / `..` / array-vs-slice rules should say `does not by itself read or capture`. That qualifier matters because subpatterns can still capture subplaces; e.g. `let (x0, ..) = x;` still captures `x.0`.

Could this subsection be adjusted along those lines?

Support:
- FLS glossary `underscore expression` / `underscore pattern` in `src/glossary.rst:4799-4811`
- Reference `type.closure.capture.precision.wildcard.reads`
- Reference `type.closure.capture.precision.wildcard.destructuring`
- Reference `type.closure.capture.precision.wildcard.fields`
- Reference `type.closure.capture.precision.slice-patterns.arrays`
- rust-lang/rust#138961

PLeVasseur: Separate point on the `Matching a slice against a slice pattern...` rule in this block: the current summary seems to compress away an important Reference detail.

Upstream treats this as a read of the slice length, but the captured place is the slice pointee rather than the wide-pointer place.

That distinction matters for cases like `&mut [T]` and `&mut &'l [u8]`, where the effective capture is `*x` / `**x`, not `x` / `*x`.

Could that pointee-vs-wide-pointer behavior be made explicit here, or backed with a brief note/example mirroring the Reference?

Support:
- Reference `type.closure.capture.precision.slice-patterns.slices`
- rust-lang/rust#138961
- rust-lang/reference#1837

PLeVasseur: Separate point on the `repr(packed)` rule in this block: this still seems narrower than the Reference.

Upstream truncates at the first field access into any `repr(packed)` struct, and explicitly says this includes fields that are currently aligned as well, for future-compatibility reasons.

Could this sentence be widened accordingly instead of limiting it to `unaligned` fields?

Support:
- Reference `type.closure.capture.precision.unaligned`

PLeVasseur (changes requested):
Thanks for pulling this material into 6.22. The remaining comments are mainly about semantic completeness and alignment with the Rust Reference so the new section lands as dependable spec text.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..06229e769561c51aa70defc2bd4a9b2854804ba9)*

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+A :t:`closure expression` :t:`[borrow]s` or :t:`moves <by move>` the :t:`capture path`, which may be truncated based on these rules:
+
+- :dp:`fls_4TESOxGpEY2h`
+  When a :t:`capture path` and an ancestor :t:`capture path` are both :t:`captured <capturing>`, the ancestor :t:`capture path` is captured with the highest :t:`capture mode` among the two :t:`[capture path]s`.
+
+- :dp:`fls_eNkZWskzznW6`
+  The :t:`capture path` is truncated at the rightmost :t:`dereference` in the :t:`capture path` if the :t:`dereference` is applied to a :t:`shared reference`.
+
+:dp:`fls_v8IFXHJnXhez`
+A :t:`place` is not captured when an :t:`underscore expression` is used to bind it.
```
PLeVasseur: Optional completeness point: upstream also states that values matched with wildcards must still be initialized.

If you want closer Reference parity in this subsection, consider adding a short note or rule for that as well.

Support:
- Reference `type.closure.capture.precision.wildcard.initialized`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..06229e769561c51aa70defc2bd4a9b2854804ba9)*

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+The entire :t:`slice` or :t:`array` is always captured even if used with :t:`underscore expression`, :t:`indexing <index expression>`, or :t:`slicing <slice>`.
+
+:dp:`fls_Vt9C9mKxHOwo`
+A :t:`place` is captured by :t:`immutable borrow` if its :t:`discriminant` is read by :t:`pattern matching`.
+
+:dp:`fls_Fs12dmznjsMf`
+Matching against a variant of an enum that has more than one variant captures the :t:`place` by :t:`immutable borrow`.
+
+:dp:`fls_7EXHdE2eOVek`
+Matching against a variant of an enum that has one variant does not capture the place, unless it is marked with :t:`attribute` ``non_exhaustive``, in which case the place is captured by :t:`immutable borrow`.
```
PLeVasseur: Separate completeness point on the discriminant-read rules: this block still seems to omit the Reference rule for uninhabited variants.

Even if all variants but the one being matched against are uninhabited, the discriminant is still read if it otherwise would be.

Could this subsection add that explicitly?

Support:
- Reference `type.closure.capture.precision.discriminants.uninhabited-variants`
- rust-lang/rust#150681

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/3cc9570466e833b66f8fe5ed0cda4ec1f67ebf4b..06229e769561c51aa70defc2bd4a9b2854804ba9)*

src/expressions.rst
tshepang: FLS does not talk about Box, so I decided to omit it... what think you about this @kirtchev-adacore

src/expressions.rst
tshepang: was thinking *an ancestor* should cover that, since ancestor can mean parent, grand parent, great grand parent, and so on... unless you don't think that is strong enough

tshepang: > This and the following rules should be list items.

it already is... see the line prior

tshepang: 85fc279f9c0429860c1d6f8d938b028af7d82ab6

src/expressions.rst
tshepang: https://github.com/rust-lang/fls/pull/698#discussion_r3132288749

src/expressions.rst
kirtchev-adacore: When the FLS was first written, only the `alloc` and `core` libraries were in scope. This was intentional, as `alloc` and `core` are OS-agnostic, and are very tightly coupled with the language semantics. `std` is another story.

We can either keep it the same way but start writing down somewhere which `std`-related items were not included in the FLS (in case we want to include them later), or start introducing `std` items into the FLS.

I think this needs a higher level discussion.

src/expressions.rst
tshepang: > array-vs-slice rules

what does this mean, a typo perhaps... the paragraph is about enums, structs, and tuples

tshepang: > should say `does not by itself read or capture`

dcd9707 adds clarity, and `fls_Vt9C9mKxHOwo` covers (maybe) this part of your concern

would that be enough

src/expressions.rst
kirtchev-adacore: To make progress on PR, I suggest @tshepang captures (no pun intended :laughing: ) the `std`-related stuff in https://github.com/rust-lang/fls/issues/700 so that we do not lose it, and remove them from this PR.

src/expressions.rst
tshepang: hm, I don't really know how to address this review... I need to get a better understanding of these things

src/changelog.rst
tshepang: I can go find where each behavior was added, but am not sure it's worth the time

maybe a good compromise is having a different heading, like "Pre-existing behavior before 1.95", or just "FLS maintenance"

src/expressions.rst
tshepang: are there std-related things in this pr @kirtchev-adacore

src/expressions.rst
tshepang: also, there are overlaps with https://github.com/rust-lang/fls/pull/698#discussion_r3115942684... maybe a topic for a meet

src/expressions.rst
kirtchev-adacore: Hmm, I do not see anything std-related. I may have been mislead by Pete's comment about `Box` and `Deref`.

src/types-and-traits.rst
tshepang: e66de62b201f7754e85af31e7a37a085601a2d62

src/expressions.rst
tshepang: I skipped that deliberately as I thought it was redundant (pedantic)... the value won't be used anyways, and if this restriction gets removed in future, it does not affect the content of the FLS.

src/expressions.rst
tshepang: > Could this sentence be widened accordingly instead of limiting it to `unaligned` fields?

I see that as additional explanation, and does not change the primary rule at all... "this is nutritious (even if you don't enjoy the taste)"

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+A :t:`place` is not :t:`captured` by being matched against a :t:`rest pattern`.
+
+:dp:`fls_RaONmCLH2KGM`
+An :t:`array` or :t:`slice` is :t:`captured` whole.
+
+:dp:`fls_Fs12dmznjsMf`
+Matching against an :t:`enum variant` of an :t:`enum` with more than one :t:`[enum variant]s` :t:`captures <capturing>` the :t:`place` by :t:`immutable borrow`.
+
+:dp:`fls_7EXHdE2eOVek`
+Matching against an :t:`enum variant` of an :t:`enum` with one :t:`enum variant` does not :t:`capture <capturing>` the :t:`place`, unless it is subject to :t:`attribute` ``non_exhaustive``, in which case the :t:`place` is captured by :t:`immutable borrow`.
```
kirtchev-adacore:
```suggestion
Matching against an :t:`enum variant` of an :t:`enum` with one :t:`enum variant` does not :t:`capture <capturing>` the :t:`place`, unless the :t:`enum` is subject to :t:`attribute` ``non_exhaustive``, in which case the :t:`place` is captured by :t:`immutable borrow`.
```
This makes it a little clearer, otherwise "it" may be interpreted as the enum variant, and not the enum.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/4ab4a37416e3fb1b47329e001f4add7bc6a9e9e2..e66de62b201f7754e85af31e7a37a085601a2d62)*

src/patterns.rst
```diff
@@ -90,6 +90,18 @@ restrictions:
   Any two :t:`[binding]s` with the same name in the two :t:`[pattern-without-alternation]s` shall have
   :t:`[unifiable type]s` and shall have the same :t:`[binding mode]s`.
 
+:dp:`fls_UfvoShNTxVD2`
+:dt:`Pattern destructuring` is where :t:`[value]s` of :t:`[struct]s`, :t:`[enum]s`, and :t:`[tuple]s` are broken into :t:`[struct field]s`, :t:`[enum variant]s`, and :t:`[tuple field]s`, respectively.
```
kirtchev-adacore:
```suggestion
:dt:`Pattern destructuring` is where :t:`[value]s` of :t:`[enum]s`, :t:`[struct]s`, and :t:`[tuple]s` are broken into :t:`[enum variant]s`, :t:`[struct field]s`, and :t:`[tuple field]s`, respectively.
```
Always alphabetize when enumerating.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/4ab4a37416e3fb1b47329e001f4add7bc6a9e9e2..e66de62b201f7754e85af31e7a37a085601a2d62)*

src/patterns.rst
```diff
@@ -90,6 +90,18 @@ restrictions:
...
+
+:dp:`fls_wAMHIDvLG0LM`
+In a pattern whose :t:`subject expression` is a :t:`struct type`, an :t:`underscore pattern` matches a single :t:`struct field`, and a :t:`rest pattern` matches the remaining fields.
+
+:dp:`fls_YuQxJCjneaW5`
+In a pattern whose :t:`subject expression` is an :t:`enum type`, an :t:`underscore pattern` matches a single :t:`enum variant`, and a :t:`rest pattern` matches the remaining variants.
+
+:dp:`fls_nrTZ4mn6IJCl`
+In a pattern whose :t:`subject expression` is a :t:`tuple type`, an :t:`underscore pattern` matches a single :t:`tuple field`, and a :t:`rest pattern` matches the remaining fields.
+
```
kirtchev-adacore:
```suggestion
:dp:`fls_YuQxJCjneaW5`
In a pattern whose :t:`subject expression` is an :t:`enum type`, an :t:`underscore pattern` matches a single :t:`enum variant`, and a :t:`rest pattern` matches the remaining variants.

:dp:`fls_wAMHIDvLG0LM`
In a pattern whose :t:`subject expression` is a :t:`struct type`, an :t:`underscore pattern` matches a single :t:`struct field`, and a :t:`rest pattern` matches the remaining fields.

:dp:`fls_nrTZ4mn6IJCl`
In a pattern whose :t:`subject expression` is a :t:`tuple type`, an :t:`underscore pattern` matches a single :t:`tuple field`, and a :t:`rest pattern` matches the remaining fields.
```
To mimic the alphabetization above.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/4ab4a37416e3fb1b47329e001f4add7bc6a9e9e2..e66de62b201f7754e85af31e7a37a085601a2d62)*

src/patterns.rst
kirtchev-adacore: I suspect that "Pattern destructuring" may need its own (short) subsection, after 5.1. Refutability.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/4ab4a37416e3fb1b47329e001f4add7bc6a9e9e2..e66de62b201f7754e85af31e7a37a085601a2d62)*

src/expressions.rst
```diff
@@ -5163,12 +5163,77 @@ within the :t:`capturing expression`, as follows:
...
+A :t:`place` is :t:`captured` by :t:`immutable borrow` if its :t:`discriminant` is read by :t:`pattern matching`.
+
+:dp:`fls_v8IFXHJnXhez`
+A :t:`place` is not :t:`captured` when an :t:`underscore pattern` is used to bind it.
+
+:dp:`fls_gujpU7p5n9Zx`
+A :t:`place` is not :t:`captured` by using a :t:`rest pattern` when destructuring :t:`[struct]s`, :t:`[tuple]s`, and :t:`[enum]s` with a single :t:`enum variant`.
+
+:dp:`fls_t8tFLUg8O83Q`
+A :t:`place` is not :t:`captured` by being matched against a :t:`rest pattern`.
```
kirtchev-adacore: I do not think this is correct. The reference [mentions](https://doc.rust-lang.org/nightly/reference/types/closure.html#r-type.closure.capture.precision.wildcard.fields) that **fields** matched against rest patterns are not captured, not places.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/4ab4a37416e3fb1b47329e001f4add7bc6a9e9e2..e66de62b201f7754e85af31e7a37a085601a2d62)*

kirtchev-adacore (dismissed):

src/expressions.rst
tshepang: 7713de9b49a58a1234c2b10832fc96bdc08ec0eb

src/patterns.rst
tshepang: 3056e1f3784b579b01e901b1b765e8e396021ac5

src/expressions.rst
tshepang: this is similar to https://github.com/rust-lang/fls/pull/698#discussion_r3181274792

src/expressions.rst
kirtchev-adacore: I read the Reference's Capture precision, and I do not think there are missing rules.

src/patterns.rst
kirtchev-adacore: Great!

src/expressions.rst
kirtchev-adacore: :+1:

kirtchev-adacore:
> This is it for now. I will have to revisit the remaining rules once the "place" business has been cleared up. I think we can find better wording/grouping of the remaining rules, as they feel somewhat randomly sprinkled.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/698/f5c48f51906cc12b5745ced6658d96ea0850552a..c0edef222591c64c0d1f484622e4d0c7565f4659)_

Just to clarify, after reading the Reference's Capture precision, I think "place" is indeed the proper term.

src/expressions.rst
kirtchev-adacore: I am OK with Tshepang's latest change. Pete however should also chime in with respect to his comment.

kirtchev-adacore: How about something along the lines of

When a :t:`capture path` and an ancestor :t:`capture path` are both :t:`captured <capturing>`, the ancestor :t:`capture path` is :t:`captured` with the highest :t:`capture mode` among the two :t:`[capture path]s`, recursively, for all such pairs of :t:`[capture path]s`.

The Reference's "recursive" note is not 100% clear because one also has to take into account pairs of paths.

**This comment is still relevant.**

src/expressions.rst
kirtchev-adacore: I propose that rule 6.22:4 is changed as follows:

Capturing is the process of saving the capture targets of a capturing expression's capturing environment, only when the capture targets need to be read.

Given the above, we have the following changes:

:dp:`fls_v8IFXHJnXhez`
A :t:`place` is not :t:`captured` when an :t:`underscore pattern` is used to bind it as the :t:`value` is not read.

:dp:`fls_gujpU7p5n9Zx`
A :t:`place` is not :t:`captured` by using a :t:`rest pattern` when :t:`destructuring <pattern destructuring>` :t:`[struct]s`, :t:`[tuple]s`, and :t:`[enum]s` with a single :t:`enum variant` as the :t:`value` is not read.

:dp:`fls_t8tFLUg8O83Q`
A :t:`field` is not :t:`captured` by being matched against a :t:`rest pattern` as the :t:`field` is not read.

:dp:`fls_HMJUXHrvOmPl`
Matching a :t:`slice` against a :t:`slice pattern`, other than one with only a single :t:`rest pattern`, :t:`captures <capturing>` the :t:`slice` by :t:`immutable borrow` as the length of the :t:`slice` is read.

:dp:`fls_Gj1znNpthHY6`
Matching an :t:`array` against a :t:`slice pattern` does not :t:`capture <capturing>` the :t:`place` as the length of the :t:`array` is fixed by its :t:`type`.

:dp:`fls_fATMTNUOHsfb`
A :t:`closure expression` :t:`captures <capturing>` the prefix of the :t:`capture path` that runs up to, but not including, the first :t:`field access expression` into a :t:`struct` that uses the :t:`attribute` :c:`repr` with modifier ``packed``.

(I removed some redundancy and relaxed the rule a bit)

**This comment is still relevant.**

src/expressions.rst
kirtchev-adacore: Here are my suggestions:

:dp:`fls_j9WyKVyOLFon`
A :dt:`place projection` is an :t:`array` :t:`index expression`, an automatic :t:`dereference`, an explicit :t:`dereference`, a :t:`field access expression`, a :t:`pattern destructuring` applied to a :t:`variable`, or a :t:`slice` :t:`index expression`.

(alphabetized, included automatic dereference)

I think that to fit "capture path" into the existing 6.22 Capturing model, we should change the rule about capture paths as follows:

:dp:`fls_rdDT7jsaOMbs`
A :dt:`capture path` is a sequence starting with a :t:`capture target` from the :t:`capturing environment` followed by zero or more :t:`[place projection]s` from that :t:`capture target`.

**This comment is still relevant.**

src/expressions.rst
kirtchev-adacore: We have the following blanket rule:

15.2:4 A variable shall be initialized before it is accessed.

where we conveniently do not explain what "accessed" means. :laughing: 

**This comment is still relevant.**

src/expressions.rst
kirtchev-adacore: The FLS does not define what an "uninhabited type" is, and it should. I propose we insert the following rules between 4.1:1 and 4.1:2:

An :t:`inhabited type` is a :t:`type` with :t:`[constructor]s` and :t:`[value]s`.

An :t:`uninhabited type` is a :t:`type` without :t:`[constructor]s` and :t:`[value]s`.

The FLS also does not define what a "constructor" is, and it should. I propose we insert a new 6.1.3 Constructor Expressions subsection, with the following singular legality rule:

A :t:`constructor` is an :t:`array expression`, a :t:`struct expression`, or a :t:`tuple expression`.

(I probably missed an expression here...)

Needless to say, all these rules will need corresponding entries in the Glossary.

Give the above, I propose the following changes:

:dp:`fls_Fs12dmznjsMf`
Matching against an :t:`enum variant` of an :t:`enum` with more than one :t:`[enum variant]s` :t:`captures <capturing>` the :t:`place` by :t:`immutable borrow` as the :t:`discriminant` is read.

:dp:`fls_7EXHdE2eOVek`
Matching against an :t:`enum variant` of an :t:`enum` with one :t:`enum variant` does not :t:`capture <capturing>` the :t:`place` as the :t:`discriminant` is not read, unless the :t:`enum` is subject to :t:`attribute` ``non_exhaustive``, in which case the :t:`place` is captured by :t:`immutable borrow`.

I do not understand what the `type.closure.capture.precision.discriminants.uninhabited-variants` Reference rule is trying to say. Is it something along the lines of "if an enum type has at least one inhabited variant, and possibly multiple uninhabited variants, then the discriminant is still read, thus capturing takes place"?

**This comment is still relevant.**

src/expressions.rst
tshepang: 98bbcfe

src/expressions.rst
tshepang: am thinking that we should not be repeating "is read" or "is not read" in these rules, as they would be implied by the rule that the captures can only happen when there is a read... a3d7be4

is it `fls_fATMTNUOHsfb` that had redundancy removed, or which one is it

kirtchev-adacore: Under what circumstances does the read happen though? If you notice, `:dp:fls_v8IFXHJnXhez` and `:dp:fls_gujpU7p5n9Zx` indicate scenarios when it looks like capturing should take place, but read does **not** take place!

It is not enough to state that capturing occurs when a read occurs because it is unclear when the actual read occurs.

src/expressions.rst
tshepang: dd47837fc3c70952b1d8794e9cba2971a59feada, though I wanted to also add definition of "automatic dereferencing", but it was very clear to me what it is, as I encountered the term "deref coercions, which looks like means the same thing. I can look further if desired.

tshepang: f355f93535ee26d822ea0403421d1b6f358f5b3b

src/expressions.rst
tshepang: @kirtchev-adacore should I create a term for it, and what would the definition be

src/expressions.rst
tshepang: it's implied that a read happens (iiuc), as that decides whether a capture happens... mentioning that a read happens is redundant

tshepang: > is it `fls_fATMTNUOHsfb` that had redundancy removed, or which one is it

you maybe missed this question

src/expressions.rst
kirtchev-adacore: For the most part, "accessed" is equivalent to "read", however unsafe code and raw pointers change the game. I think it is possible to play pointer arithmetic games and read chunks of a variable, such as fields, but in a low-level fashion.

I guess 15.2:4 can be rephrased as
```
A variable shall be initialized before either the variable or any of its fields are read.
```

src/expressions.rst
kirtchev-adacore: > > is it `fls_fATMTNUOHsfb` that had redundancy removed, or which one is it
> 
> you maybe missed this question

I am confused, what redundancy are you referring to?

kirtchev-adacore: > it's implied that a read happens (iiuc), as that decides whether a capture happens... mentioning that a read happens is redundant

I have the vague memory that perhaps @PLeVasseur suggested that the rules should indicate when a read happens, but I can't find the comment...

tshepang: > I am confused, what redundancy are you referring to?

it's in this comment https://github.com/rust-lang/fls/pull/698#discussion_r3217129461

kirtchev-adacore: Hmm, I can't find where the redundancy was, even after comparing the two suggestions. :shrug:

src/expressions.rst
tshepang: is that change worth making, seeing that "access" is used throughout the text (like "a field access expression is an expressions that accesses a field of a value")

src/expressions.rst
tshepang: @kirtchev-adacore could these suggestions be a separate issue(s) instead... feels like it needs further refinement, and am not confident on the suggestion

src/expressions.rst
kirtchev-adacore: Not worth it then.

src/expressions.rst
kirtchev-adacore: A separate issue sounds good!

tshepang: https://github.com/rust-lang/fls/issues/705

src/expressions.rst
kirtchev-adacore: Looks good!

src/changelog.rst
kirtchev-adacore: @tshepang So far we have used the following headings:
```rst
FLS fixes/improvements
FLS corrections
```
I think those should be changed to "FLS maintenance" (as it is more general), and you should use that heading to split the pre-release changes.

tshepang:
rebased due to divergence from `main`

src/changelog.rst
tshepang: done through [a rebase](https://triagebot.infra.rust-lang.org/gh-range-diff/rust-lang/fls/4ab4a37416e3fb1b47329e001f4add7bc6a9e9e2..529319ec97909d1430453eb614755ab66368b06d/3977e597187d55140d57c75227a116b6afc7a852..c14483ebb47f763a5d15c9c94d2533ac27425a0a)

kirtchev-adacore (approved):

PR #707, opened 2026-07-01
[1.95.0] Document provenance and its use in const eval

kirtchev-adacore:
The information has been taken from
- [core::ptr Provenance](https://doc.rust-lang.org/core/ptr/index.html#provenance)
- [reference#2091](https://github.com/rust-lang/reference/pull/2091)
- [reference#2139](https://github.com/rust-lang/reference/pull/2138)

### DIFF
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -163,6 +163,9 @@ Language changes in Rust 1.95.0
   - The target is outside the scope of the FLS.
 
 - `const-eval: be more consistent in the behavior of padding during typed copies <https://github.com/rust-lang/rust/pull/148967>`_
+
+  - TODO: Fill out once we settle on the rules.
+
 - `Const blocks are no longer evaluated to determine if expressions involving fallible operations can implicitly be constant-promoted <https://github.com/rust-lang/rust/pull/150557>`_
 
   - This implementation-specifc behavior is outside the scope of the FLS.
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -49,7 +49,7 @@ Language changes in Rust 1.95.0
   - :t:`match arm guard operand`
   - :t:`match arm guard scrutinee`
 
-  Updated glossary entries:
+  Changed glossary entries:
 
   - :t:`let binding`
 
@@ -164,7 +164,44 @@ Language changes in Rust 1.95.0
 
 - `const-eval: be more consistent in the behavior of padding during typed copies <https://github.com/rust-lang/rust/pull/148967>`_
 
-  - TODO: Fill out once we settle on the rules.
+  New glossary entries:
+
+  - :t:`derived pointer`
+  - :t:`original pointer`
+  - :t:`provenance`
+  - :t:`well-formed pointer`
+
+  Changed glossary entries:
+  
+  - :t:`pointer type`
+  - :t:`raw pointer`
+
+  New paragraphs:
+
+  - :p:`fls_3qI8FXMsyk0f`
+  - :p:`fls_c2Guy3fPYaUV`
+  - :p:`fls_jriT46yWgIR0`
+  - :p:`fls_VWUlxTy0QF9d`
+  - :p:`fls_kaPNJ7iIHPro`
+  - :p:`fls_5MkKtNL9oCsL`
+  - :p:`fls_1NJhTBN1D2qv`
+  - :p:`fls_wnJmQYT7iKQf`
+  - :p:`fls_ffh8mAkebORJ`
+  - :p:`fls_c3DaCLQEBpYQ`
+  - :p:`fls_l1FOH8zt0XRZ`
+  - :p:`fls_37oocZVDne5Y`
+
+  Changed paragraphs:
+
+  - :p:`fls_hrum767l6dte`
+  - :p:`fls_k6ues2936pjq`
+  - :p:`fls_jd1inwz7ulyw`
+
+  Removed paragraphs:
+
+  - :p:`fls_deuo1pn8cjd6`
+
+  New section: :ref:`fls_Dqk4eIvxHloY`
 
 - `Const blocks are no longer evaluated to determine if expressions involving fallible operations can implicitly be constant-promoted <https://github.com/rust-lang/rust/pull/150557>`_
 
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -1159,7 +1159,7 @@ A :dt:`derive macro` is a :t:`procedural macro` that consumes a stream of
 derived pointer
 ^^^^^^^^^^^^^^^
 
-A :dt:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`, copying a :t:`pointer`, loading a stored :t:`pointer`, performing :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
+A :dt:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`, copying a :t:`pointer`, reading a stored :t:`pointer`, performing :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
 
 destruction
 ^^^^^^^^^^^
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -3488,7 +3488,7 @@ a Rust program.
 provenance
 ^^^^^^^^^^
 
-:dt:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the accesses may read from or write to memory.
+:dt:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the accesses may read from or write to said memory locations.
 
 public visibility
 ^^^^^^^^^^^^^^^^^
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -1156,6 +1156,11 @@ A :dt:`derive macro` is a :t:`procedural macro` that consumes a stream of
 :t:`[token]s` and produces a stream of tokens, and is invoked via attribute
 :c:`derive`.
 
+derived indirection value
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A :dt:`derived indirection value` is an :t:`indirection value` obtained by performing address or pointer arithmetic on another :t:`indirection value`.
+
 destruction
 ^^^^^^^^^^^
 
@@ -2193,6 +2198,11 @@ indirection type
 An :dt:`indirection type` is a :t:`type` whose :t:`[value]s` refer to memory
 locations.
 
+indirection value
+^^^^^^^^^^^^^^^^^
+
+An :dt:`indirection value` is either a :t:`raw pointer` or a :t:`reference`.
+
 inert attribute
 ^^^^^^^^^^^^^^^
 
@@ -3186,13 +3196,17 @@ opt-out trait bound
 An :dt:`opt-out trait bound` is a :t:`trait bound` with :s:`Punctuation` ``?``
 that nullifies an implicitly added :t:`trait bound`.
 
-
 or-pattern
 ^^^^^^^^^^
 
 An :dt:`or-pattern` is a :t:`pattern` that matches on one of two or more :t:`[pattern-without-alternation]s` and or-s them using character 0x7C (vertical line, i.e. ``|``).
 
 See :s:`Pattern`.
+
+original indirection value
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+An :dt:`original indirection value` is an :t:`indirection value` created via allocation.
 
 outer attribute
 ^^^^^^^^^^^^^^^
@@ -3476,6 +3490,11 @@ program entry point
 A :dt:`program entry point` is a :t:`function` that is invoked at the start of
 a Rust program.
 
+provenance
+^^^^^^^^^^
+
+:dt:`Provenance` is an optional property of :t:`[indirection value]s` that restricts the addresses an :t:`indirection value` may point to, the timespan during which the :t:`indirection value` may point to those addresses, and whether the :t:`indirection value` can read from and write to those addresses.
+
 public visibility
 ^^^^^^^^^^^^^^^^^
 
@@ -3659,7 +3678,7 @@ See :s:`RawCStringLiteral`.
 raw pointer
 ^^^^^^^^^^^
 
-A :dt:`raw pointer` is a pointer of a :t:`raw pointer type`.
+A :dt:`raw pointer` is a :t:`value` of a :t:`raw pointer type`.
 
 raw pointer type
 ^^^^^^^^^^^^^^^^
@@ -5185,6 +5204,11 @@ A :dt:`weak keyword` is a :t:`keyword` whose special meaning depends on the
 context.
 
 See :s:`WeakKeyword`.
+
+well-formed indirection value
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A :dt:`well-formed indirection value` is an :t:`indirection value` with :t:`provenance`, where all bytes that comprise the :t:`indirection value` are initialized, correctly ordered, and are fragments of the same :t:`original indirection value`.
 
 where clause
 ^^^^^^^^^^^^
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -1159,7 +1159,7 @@ A :dt:`derive macro` is a :t:`procedural macro` that consumes a stream of
 derived pointer
 ^^^^^^^^^^^^^^^
 
-A :dt:`derived pointer` is a :t:`pointer` obtained by performing address or pointer arithmetic on another :t:`pointer`.
+A :dt:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`, copying a :t:`pointer`, loading a stored :t:`pointer`, performing :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
 
 destruction
 ^^^^^^^^^^^
@@ -3420,7 +3420,7 @@ A :dt:`pointer` is a :t:`value` of a :t:`pointer type`.
 pointer type
 ^^^^^^^^^^^^
 
-A :dt:`pointer type` is a :t:`type` whose :t:`[value]s` indicate memory locations.
+A :dt:`pointer type` is either a :t:`raw pointer type` or a :t:`reference type`.
 
 positional register argument
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -3488,7 +3488,7 @@ a Rust program.
 provenance
 ^^^^^^^^^^
 
-:dt:`Provenance` is an optional property of :t:`[pointer]s` that restricts the addresses a :t:`pointer` may point to, the timespan during which the :t:`pointer` may point to those addresses, and whether the :t:`pointer` can read from and write to those addresses.
+:dt:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the accesses may read from or write to memory.
 
 public visibility
 ^^^^^^^^^^^^^^^^^
@@ -5203,7 +5203,7 @@ See :s:`WeakKeyword`.
 well-formed pointer
 ^^^^^^^^^^^^^^^^^^^
 
-A :dt:`well-formed pointer` is a :t:`pointer` with :t:`provenance`, where all bytes that comprise the :t:`pointer` are initialized, correctly ordered, and are fragments of the same :t:`original pointer`.
+A :dt:`well-formed pointer` is a :t:`pointer` where either no byte of the :t:`pointer` carries :t:`provenance`, or every byte of the :t:`pointer` is the corresponding byte of a single :t:`pointer` with :t:`provenance`.
 
 where clause
 ^^^^^^^^^^^^
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -1156,10 +1156,10 @@ A :dt:`derive macro` is a :t:`procedural macro` that consumes a stream of
 :t:`[token]s` and produces a stream of tokens, and is invoked via attribute
 :c:`derive`.
 
-derived indirection value
-^^^^^^^^^^^^^^^^^^^^^^^^^
+derived pointer
+^^^^^^^^^^^^^^^
 
-A :dt:`derived indirection value` is an :t:`indirection value` obtained by performing address or pointer arithmetic on another :t:`indirection value`.
+A :dt:`derived pointer` is a :t:`pointer` obtained by performing address or pointer arithmetic on another :t:`pointer`.
 
 destruction
 ^^^^^^^^^^^
@@ -2198,11 +2198,6 @@ indirection type
 An :dt:`indirection type` is a :t:`type` whose :t:`[value]s` refer to memory
 locations.
 
-indirection value
-^^^^^^^^^^^^^^^^^
-
-An :dt:`indirection value` is either a :t:`raw pointer` or a :t:`reference`.
-
 inert attribute
 ^^^^^^^^^^^^^^^
 
@@ -3203,10 +3198,10 @@ An :dt:`or-pattern` is a :t:`pattern` that matches on one of two or more :t:`[pa
 
 See :s:`Pattern`.
 
-original indirection value
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+original pointer
+^^^^^^^^^^^^^^^^
 
-An :dt:`original indirection value` is an :t:`indirection value` created via allocation.
+An :dt:`original pointer` is a :t:`pointer` created via allocation.
 
 outer attribute
 ^^^^^^^^^^^^^^^
@@ -3425,7 +3420,7 @@ A :dt:`pointer` is a :t:`value` of a :t:`pointer type`.
 pointer type
 ^^^^^^^^^^^^
 
-A :dt:`pointer type` is a :t:`type` whose values indicate memory locations.
+A :dt:`pointer type` is a :t:`type` whose :t:`[value]s` indicate memory locations.
 
 positional register argument
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -3493,7 +3488,7 @@ a Rust program.
 provenance
 ^^^^^^^^^^
 
-:dt:`Provenance` is an optional property of :t:`[indirection value]s` that restricts the addresses an :t:`indirection value` may point to, the timespan during which the :t:`indirection value` may point to those addresses, and whether the :t:`indirection value` can read from and write to those addresses.
+:dt:`Provenance` is an optional property of :t:`[pointer]s` that restricts the addresses a :t:`pointer` may point to, the timespan during which the :t:`pointer` may point to those addresses, and whether the :t:`pointer` can read from and write to those addresses.
 
 public visibility
 ^^^^^^^^^^^^^^^^^
@@ -5205,10 +5200,10 @@ context.
 
 See :s:`WeakKeyword`.
 
-well-formed indirection value
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+well-formed pointer
+^^^^^^^^^^^^^^^^^^^
 
-A :dt:`well-formed indirection value` is an :t:`indirection value` with :t:`provenance`, where all bytes that comprise the :t:`indirection value` are initialized, correctly ordered, and are fragments of the same :t:`original indirection value`.
+A :dt:`well-formed pointer` is a :t:`pointer` with :t:`provenance`, where all bytes that comprise the :t:`pointer` are initialized, correctly ordered, and are fragments of the same :t:`original pointer`.
 
 where clause
 ^^^^^^^^^^^^
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -942,7 +942,7 @@ Indirection Types
 .. rubric:: Legality Rules
 
 :dp:`fls_3qI8FXMsyk0f`
-A :t:`pointer type` is a :t:`type` whose :t:`[value]s` indicate memory locations.
+A :t:`pointer type` is either a :t:`raw pointer type` or a :t:`reference type`.
 
 .. _fls_xztr1kebz8bo:
 
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -939,6 +939,11 @@ A :t:`function item type` implements the :std:`core::ops::Fn` :t:`trait`, the :s
 Indirection Types
 -----------------
 
+.. rubric:: Legality Rules
+
+:dp:`fls_3qI8FXMsyk0f`
+A :t:`pointer type` is a :t:`type` whose :t:`[value]s` indicate memory locations.
+
 .. _fls_xztr1kebz8bo:
 
 Function Pointer Types
--- a/src/types-and-traits.rst
+++ b/src/types-and-traits.rst
@@ -1028,13 +1028,14 @@ A :t:`mutable raw pointer type` is a :t:`raw pointer type` subject to
 An :t:`immutable raw pointer type` is a :t:`raw pointer type` subject to
 :t:`keyword` ``const``.
 
+:dp:`fls_c2Guy3fPYaUV`
+A :t:`raw pointer` is a :t:`value` of a :t:`raw pointer type`.
+
 :dp:`fls_hrum767l6dte`
-Comparing two :t:`[value]s` of :t:`[raw pointer type]s` compares the addresses
-of the :t:`[value]s`.
+Comparing two :t:`[raw pointer]s` compares the addresses of the :t:`[raw pointer]s`.
 
 :dp:`fls_k6ues2936pjq`
-Comparing a :t:`value` of a :t:`raw pointer type` to a :t:`value` of a
-:t:`dynamically sized type` compares the data being pointed to.
+Comparing a :t:`raw pointer` to a :t:`value` of a :t:`dynamically sized type` compares the data being pointed to.
 
 .. rubric:: Examples
 
--- a/src/values.rst
+++ b/src/values.rst
@@ -141,7 +141,7 @@ The :t:`expression` of a :t:`constant initializer` shall be a
 :t:`constant expression`.
 
 :dp:`fls_l1FOH8zt0XRZ`
-If the :t:`value` produced by evaluating a :t:`constant initializer` is or contains a :t:`pointer`, then each such :t:`pointer` shall be a :t:`well-formed pointer`.
+If the :t:`value` produced by evaluating a :t:`constant initializer` is or contains a :t:`pointer`, then each such :t:`pointer` shall be a non-dangling :t:`well-formed pointer`.
 
 .. rubric:: Dynamic Semantics
 
@@ -221,7 +221,7 @@ The :t:`expression` of a :t:`static initializer` shall be a
 :t:`constant expression`.
 
 :dp:`fls_37oocZVDne5Y`
-If the :t:`value` produced by evaluating a :t:`static initializer` is or contains a :t:`pointer`, then each such :t:`pointer` shall be a :t:`well-formed pointer`.
+If the :t:`value` produced by evaluating a :t:`static initializer` is or contains a :t:`pointer`, then each such :t:`pointer` shall be a non-dangling :t:`well-formed pointer`.
 
 :dp:`fls_8dcldbvu7lav`
 A use of a :t:`static` is a :t:`place expression` referring to the unique
--- a/src/values.rst
+++ b/src/values.rst
@@ -76,7 +76,7 @@ A :t:`pointer` is a :t:`value` of a :t:`pointer type`.
 An :t:`original pointer` is a :t:`pointer` created via allocation.
 
 :dp:`fls_kaPNJ7iIHPro`
-A :t:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`, copying a :t:`pointer`, loading a stored :t:`pointer`, performing :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
+A :t:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`, copying a :t:`pointer`, reading a stored :t:`pointer`, performing :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
 
 :dp:`fls_5MkKtNL9oCsL`
 :t:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the accesses may read from or write to said memory locations.
--- a/src/values.rst
+++ b/src/values.rst
@@ -42,15 +42,6 @@ Two :t:`[value]s` :t:`overlap` when
 * :dp:`fls_eoak5mdl6ma`
   Both :t:`[value]s` are elements of the same :t:`array`.
 
-:dp:`fls_jriT46yWgIR0`
-A :t:`pointer` is a :t:`value` of a :t:`pointer type`.
-
-:dp:`fls_VWUlxTy0QF9d`
-An :t:`original pointer` is a :t:`pointer` created via allocation.
-
-:dp:`fls_kaPNJ7iIHPro`
-A :t:`derived pointer` is a :t:`pointer` obtained by performing address or pointer arithmetic on another :t:`pointer`.
-
 .. rubric:: Undefined Behavior
 
 :dp:`fls_6lg0oaaopc26`
@@ -70,6 +61,39 @@ than the architectures maximum :c:`usize` value.
 It is undefined behavior to create an :t:`allocated object` with :t:`memory
 size` ``size`` where ``size`` is greater than the architectures maximum
 :c:`isize` value.
+
+.. _fls_Dqk4eIvxHloY:
+
+Provenance
+----------
+
+.. rubric:: Legality Rules
+
+:dp:`fls_jriT46yWgIR0`
+A :t:`pointer` is a :t:`value` of a :t:`pointer type`.
+
+:dp:`fls_VWUlxTy0QF9d`
+An :t:`original pointer` is a :t:`pointer` created via allocation.
+
+:dp:`fls_kaPNJ7iIHPro`
+A :t:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`, copying a :t:`pointer`, loading a stored :t:`pointer`, performing :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
+
+:dp:`fls_5MkKtNL9oCsL`
+:t:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the accesses may read from or write to memory.
+
+:dp:`fls_1NJhTBN1D2qv`
+An :t:`original pointer` carries :t:`provenance` over all or part of the :t:`allocated object` it was created from.
+
+:dp:`fls_wnJmQYT7iKQf`
+A :t:`derived pointer` inherits the :t:`provenance` of the :t:`pointer` it was created from, if any.
+
+:dp:`fls_ffh8mAkebORJ`
+A :t:`well-formed pointer` is a :t:`pointer` where either no byte of the :t:`pointer` carries :t:`provenance`, or every byte of the :t:`pointer` is the corresponding byte of a single :t:`pointer` with :t:`provenance`.
+
+.. rubric:: Undefined Behavior
+
+:dp:`fls_c3DaCLQEBpYQ`
+It is undefined behavior to access memory through a :t:`pointer` that does not have :t:`provenance` permitting the access.
 
 .. _fls_ixjc5jaamx84:
 
@@ -116,12 +140,8 @@ A :t:`constant` shall have a :t:`constant initializer`, unless it is an
 The :t:`expression` of a :t:`constant initializer` shall be a
 :t:`constant expression`.
 
-:dp:`fls_deuo1pn8cjd6`
-The :t:`value` of a :t:`constant` is determined by evaluating its
-:t:`constant initializer`.
-
 :dp:`fls_l1FOH8zt0XRZ`
-If the :t:`value` produced by evaluating a :t:`constant initializer` is a :t:`pointer`, then the :t:`pointer` shall be a :t:`well-formed pointer`.
+If the :t:`value` produced by evaluating a :t:`constant initializer` is or contains a :t:`pointer`, then each such :t:`pointer` shall be a :t:`well-formed pointer`.
 
 .. rubric:: Dynamic Semantics
 
@@ -200,11 +220,8 @@ A :t:`static` shall have a :t:`static initializer`, unless it is an
 The :t:`expression` of a :t:`static initializer` shall be a
 :t:`constant expression`.
 
-:dp:`fls_41zoQ54MiLMm`
-The :t:`static` is determined by evaluating its :t:`static initializer`.
-
 :dp:`fls_37oocZVDne5Y`
-If the :t:`value` produced by evaluating a :t:`static initializer` is a :t:`pointer`, then the :t:`pointer` shall be a :t:`well-formed pointer`.
+If the :t:`value` produced by evaluating a :t:`static initializer` is or contains a :t:`pointer`, then each such :t:`pointer` shall be a :t:`well-formed pointer`.
 
 :dp:`fls_8dcldbvu7lav`
 A use of a :t:`static` is a :t:`place expression` referring to the unique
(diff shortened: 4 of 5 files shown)


src/values.rst
```diff
@@ -108,9 +117,12 @@ The :t:`expression` of a :t:`constant initializer` shall be a
 :t:`constant expression`.
 
 :dp:`fls_deuo1pn8cjd6`
-The value of a :t:`constant` is determined by evaluating its
+The :t:`value` of a :t:`constant` is determined by evaluating its
 :t:`constant initializer`.
 
+:dp:`fls_l1FOH8zt0XRZ`
+If the final :t:`value` produced by evaluating a :t:`constant initializer` denotes an :t:`indirection value`, then the final :t:`value` shall be a :t:`well-formed indirection value`.
```
tshepang:
```suggestion
If the :t:`value` produced by evaluating a :t:`constant initializer` denotes an :t:`indirection value`, then the :t:`value` shall be a :t:`well-formed indirection value`.
```
"final" sounds to me like an implementation detail not worth mentioning... "produced" sounds final enough to me

see https://github.com/rust-lang/reference/pull/2138#discussion_r2773988737

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..812a13fdd8bf4a31d8857be78ff46ba35b3323d9)*

src/values.rst
```diff
@@ -188,6 +200,12 @@ A :t:`static` shall have a :t:`static initializer`, unless it is an
 The :t:`expression` of a :t:`static initializer` shall be a
 :t:`constant expression`.
 
+:dp:`fls_41zoQ54MiLMm`
+The :t:`value` of a :t:`static` is determined by evaluating its :t:`static initializer`.
```
tshepang:
```suggestion
A :t:`static` is determined by evaluating its :t:`static initializer`.
```
does this work, given that "A static is a value that is associated with a specific memory location" (else the expansion would be "A value of a value that is associated...")

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..812a13fdd8bf4a31d8857be78ff46ba35b3323d9)*

src/values.rst
```diff
@@ -188,6 +200,12 @@ A :t:`static` shall have a :t:`static initializer`, unless it is an
...
+
+:dp:`fls_37oocZVDne5Y`
+If the final :t:`value` produced by evaluating a :t:`static initializer` denotes an :t:`indirection value`, then the final :t:`value` shall be a :t:`well-formed indirection value`.
```
tshepang: see https://github.com/rust-lang/fls/pull/707/changes#r3504558369

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..812a13fdd8bf4a31d8857be78ff46ba35b3323d9)*

src/values.rst
```diff
@@ -306,3 +324,39 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
...
+:dp:`fls_ffh8mAkebORJ`
+A :t:`well-formed indirection value` is an :t:`indirection value` with :t:`provenance`, where all bytes that comprise the :t:`indirection value` are part of the same group.
+
+.. rubric:: Undefined Behavior
+
+:dp:`fls_c3DaCLQEBpYQ`
+It is undefined behavior to access memory through an :t:`indirection value` which does not have :t:`provenance` over that memory.
+
+:dp:`fls_MwDoxVPRZCqm`
+It is undefined behavior to offset an :t:`indirection value` across the bounds of the memory block it has :t:`provenance` over.
```
tshepang:
```suggestion
It is undefined behavior to offset an :t:`indirection value` beyond the bounds of the memory block it has :t:`provenance` over.
```
maybe more clear

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..812a13fdd8bf4a31d8857be78ff46ba35b3323d9)*

src/values.rst
kirtchev-adacore: Makes sense, fixed.

src/values.rst
kirtchev-adacore: Sounds good, fixed.

src/values.rst
kirtchev-adacore: Fixed.

src/values.rst
kirtchev-adacore: It is clearer, fixed.

src/values.rst
```diff
@@ -306,3 +324,39 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
...
+:t:`Provenance` is an optional property of :t:`[indirection value]s` that grants the following permissions:
+
+* :dp:`fls_zTkHLtcNnGhJ`
+  :t:`Provenance` restricts the addresses an :t:`indirection value` may point to within a block of memory.
+
+* :dp:`fls_V43OBrVuME8U`
+  :t:`Provenance` restricts the timespan during which an :t:`indirection value` may point to a block of memory.
+
+* :dp:`fls_cIfikHa7UTa8`
+  :t:`Provenance` restricts the read/write access an :t:`indirection value` has over a block of memory.
```
tshepang: it is still readable if we compress this definition, which also has these advantages
- does not leave the term in the glossary so bare, "Provenance is an optional property of indirection values that grants certain permissions"
- avoids "block of memory", which implies that the memory accessed is contiguous, a guarantee we do not need to make 
```suggestion
:dp:`fls_5MkKtNL9oCsL`
:t:`Provenance` is an optional property of :t:`[indirection value]s` that restricts the addresses an :t:`indirection value` may point to, the timespan during which the :t:`indirection value` may point to those addresses, and whether the :t:`indirection value` can write to those addresses.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..ecca0f03d6419d389b6e18555b1bee0a88e0b6b2)*

kirtchev-adacore: Note that "block of memory" is really "chunk of memory", but that does not sound very formal.

I don't think you can allocate multiple chunks of memory that a **single** pointer can point to in a **single** allocation call, even if you implement your own allocator. One of the core features of provenance is to stop you from poking into memory you are not supposed to look at, at least via undefined behavior, so we need to talk about some kind of "range of memory", whether that is a block or a chunk, or something else.

tshepang: my suggestion does not conflict with that explanation (and the source documentation says it being contiguous may not hold in future)

src/values.rst
```diff
@@ -108,9 +117,12 @@ The :t:`expression` of a :t:`constant initializer` shall be a
 :t:`constant expression`.
 
 :dp:`fls_deuo1pn8cjd6`
-The value of a :t:`constant` is determined by evaluating its
+The :t:`value` of a :t:`constant` is determined by evaluating its
 :t:`constant initializer`.
 
+:dp:`fls_l1FOH8zt0XRZ`
+If the :t:`value` produced by evaluating a :t:`constant initializer` denotes an :t:`indirection value`, then the final :t:`value` shall be a :t:`well-formed indirection value`.
```
tshepang:
```suggestion
If the :t:`value` produced by evaluating a :t:`constant initializer` denotes an :t:`indirection value`, then the :t:`value` shall be a :t:`well-formed indirection value`.
```
note this was in the previous suggestion, so maybe your workflow can be: accept the suggestion (via github ui), then squash if needed (locally)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..ecca0f03d6419d389b6e18555b1bee0a88e0b6b2)*

src/values.rst
```diff
@@ -188,6 +200,12 @@ A :t:`static` shall have a :t:`static initializer`, unless it is an
 The :t:`expression` of a :t:`static initializer` shall be a
 :t:`constant expression`.
 
+:dp:`fls_41zoQ54MiLMm`
+The :t:`static` is determined by evaluating its :t:`static initializer`.
+
+:dp:`fls_37oocZVDne5Y`
+If the :t:`value` produced by evaluating a :t:`static initializer` denotes an :t:`indirection value`, then the final :t:`value` shall be a :t:`well-formed indirection value`.
```
tshepang:
```suggestion
If the :t:`value` produced by evaluating a :t:`static initializer` denotes an :t:`indirection value`, then the :t:`value` shall be a :t:`well-formed indirection value`.
```


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..ecca0f03d6419d389b6e18555b1bee0a88e0b6b2)*

src/values.rst
kirtchev-adacore: I updated the definition to also include reading.

src/values.rst
kirtchev-adacore: Fixed.

src/values.rst
kirtchev-adacore: Fixed.

src/glossary.rst
```diff
@@ -5141,6 +5160,11 @@ context.
 
 See :s:`WeakKeyword`.
 
+well-formed indirection value
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A :dt:`well-formed indirection value` is an :t:`indirection value` with :t:`provenance`, where all bytes that comprise the :t:`indirection value` are part of the same group.
```
tshepang: the definition feels incomplete, but I don't know yet how to fill it... maybe inspiration can be taken from [a review from my old pr](https://github.com/rust-lang/fls/pull/695/changes#r3481039037)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..21bb6de29dd62a21d6af72afbf60ee74adf4c82c)*

src/glossary.rst
```diff
@@ -2193,6 +2198,11 @@ indirection type
 An :dt:`indirection type` is a :t:`type` whose :t:`[value]s` refer to memory
 locations.
 
+indirection value
+^^^^^^^^^^^^^^^^^
+
+An :dt:`indirection value` is either a :t:`raw pointer` or a :t:`reference`.
```
tshepang: section 4.7, Indirection Types, includes function pointers... should they be included here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/3c73fb2263ce3ffc9f7d7592a910d42372cad001..21bb6de29dd62a21d6af72afbf60ee74adf4c82c)*

kirtchev-adacore: No, as [indicated](https://rust-lang.zulipchat.com/#narrow/channel/520710-t-lang.2Ffls/topic/.E2.9C.94.20Provenance/with/607336254) by bjorn3.

src/glossary.rst
kirtchev-adacore: Maybe
```rst
... bytes that comprise the :t:`indirection value` are initialized and are part
of the same group.
```
Basically we want to say that an indirection value does not (somehow) contain the bytes of something else.

tshepang: is this about "group of bytes"

kirtchev-adacore: Right.

I think what the Reference PRs and `core::ptr` are trying to say is that you can obtain a value by doing sketchy things, such as memory transmutation. If you plan on using that value as an indirection value with provenance, then the value better have "correct bytes", what 2091 calls "bytes must be fragments of the same original pointer value in the correct order" and 2138 calls "part of an adjacent group of correctly-ordered bytes that form an entire pointer".

kirtchev-adacore: I updated the definition of "well-formed indirection value".

kirtchev-adacore: I updated the PR with the terminology suggested by Ralph.

src/values.rst
```diff
@@ -306,3 +324,30 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
...
+:t:`Provenance` is an optional property of :t:`[pointer]s` that restricts the addresses a :t:`pointer` may point to, the timespan during which the :t:`pointer` may point to those addresses, and whether the :t:`pointer` can read from and write to those addresses.
+
+:dp:`fls_1NJhTBN1D2qv`
+An :t:`original pointer` always carries :t:`provenance`.
+
+:dp:`fls_wnJmQYT7iKQf`
+A :t:`derived pointer` inherits the :t:`provenance` of the :t:`pointer` it was created from, if any.
+
+:dp:`fls_ffh8mAkebORJ`
+A :t:`well-formed pointer` is a :t:`pointer` with :t:`provenance`, where all bytes that comprise the :t:`pointer` are initialized, correctly ordered, and are fragments of the same :t:`original pointer`.
```
PLeVasseur: The pointer terminology reads cleaner than "indirection value" since it now matches "how the Project talks about Rust".

On the definition itself, three things seem to remain, one of which is the big one.

1. The definition still opens with "a pointer *with* provenance", and the const/static rules require well-formedness, so provenance-free pointer constants are still ruled illegal even though rustc accepts both and neither carries provenance. The fragment restriction from rust-lang/rust#144081 / rust-lang/rust#148259 applies to bytes that carry provenance; pointers whose bytes carry none are fine.

   ```rust
   const P: *const u8 = 5 as *const u8;
   const N: *const u8 = core::ptr::null();
   ```

2. "Fragments of the same :t:`original pointer`" over-constrains the source. The fragment mechanism reassembles a pointer *value*, whatever its derivation — nothing requires it to be original in this PR's sense ("created via allocation"). Concretely, `const P: *const i32 = &1 as *const i32;` is accepted, and the value's bytes come from a *derived* (cast) pointer. "A single pointer with provenance" is the condition that matches the compiler.

3. "Correctly ordered" doesn't quite pin position and completeness, since an ordered-but-offset subset of a larger pointer's bytes could be read as satisfying it. "Every byte is the corresponding byte" helps close that.

Putting those together, for the chapter paragraph:

```suggestion
A :t:`well-formed pointer` is a :t:`pointer` where either no byte of the
:t:`pointer` carries :t:`provenance`, or every byte of the :t:`pointer` is the
corresponding byte of a single :t:`pointer` with :t:`provenance`.
```

(and the same wording with `:dt:` on the defined term for the glossary entry).

Regarding "initialized": to me this seems fine either way. It's already implied as creating any pointer value from uninitialized memory is UB via `fls_6lg0oaaopc26`, and its union exception can't apply to a raw pointer or reference. So we could keep "initialized" for emphasis or dropped for brevity without changing what's legal.

src/values.rst
```diff
@@ -108,9 +117,12 @@ The :t:`expression` of a :t:`constant initializer` shall be a
 :t:`constant expression`.
 
 :dp:`fls_deuo1pn8cjd6`
-The value of a :t:`constant` is determined by evaluating its
+The :t:`value` of a :t:`constant` is determined by evaluating its
 :t:`constant initializer`.
 
+:dp:`fls_l1FOH8zt0XRZ`
+If the :t:`value` produced by evaluating a :t:`constant initializer` is a :t:`pointer`, then the :t:`pointer` shall be a :t:`well-formed pointer`.
```
PLeVasseur: "If the value ... is a :t:`pointer`" only binds when the constant's value is a pointer at top level. The compiler's validation of final values recurses through the whole value — the fragment check lands there (see the interning/validation discussion in rust-lang/rust#144081 and rust-lang/rust#148259) — so pointers inside aggregates are checked too. These are in scope for the compiler but not covered by the current wording:

```rust
const C: (u32, &u32) = (0, &1);
const D: [*const u8; 2] = [&2u8, &3u8];
```

Suggest quantifying over contained pointers. For the constant rule:

```suggestion
If the :t:`value` produced by evaluating a :t:`constant initializer` is or
contains a :t:`pointer`, then each such :t:`pointer` shall be a
:t:`well-formed pointer`.
```

src/values.rst
```diff
@@ -188,6 +200,12 @@ A :t:`static` shall have a :t:`static initializer`, unless it is an
 The :t:`expression` of a :t:`static initializer` shall be a
 :t:`constant expression`.
 
+:dp:`fls_41zoQ54MiLMm`
+The :t:`static` is determined by evaluating its :t:`static initializer`.
+
+:dp:`fls_37oocZVDne5Y`
+If the :t:`value` produced by evaluating a :t:`static initializer` is a :t:`pointer`, then the :t:`pointer` shall be a :t:`well-formed pointer`.
```
PLeVasseur: The same aggregate-coverage issue applies to `fls_37oocZVDne5Y`: the compiler validates final values recursively, so pointers inside aggregates are checked too. Suggest quantifying over contained pointers for the static rule:

```suggestion
If the :t:`value` produced by evaluating a :t:`static initializer` is or
contains a :t:`pointer`, then each such :t:`pointer` shall be a
:t:`well-formed pointer`.
```

src/values.rst
```diff
@@ -306,3 +324,30 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
...
+:dp:`fls_ffh8mAkebORJ`
+A :t:`well-formed pointer` is a :t:`pointer` with :t:`provenance`, where all bytes that comprise the :t:`pointer` are initialized, correctly ordered, and are fragments of the same :t:`original pointer`.
+
+.. rubric:: Undefined Behavior
+
+:dp:`fls_c3DaCLQEBpYQ`
+It is undefined behavior to access memory through a :t:`pointer` which does not have :t:`provenance` over that memory.
+
+:dp:`fls_MwDoxVPRZCqm`
+It is undefined behavior to offset a :t:`pointer` beyond the bounds of the memory block it has :t:`provenance` over.
```
PLeVasseur: Hrm, I can foresee two problems possibly. Right fix here might be deletion?

1. This does not hold for wrapping arithmetic. `core::ptr` is explicit that `wrapping_offset`/`wrapping_add` may leave the allocation with no UB; only the in-bounds family (`offset`, `add`, `sub`) carries the bounds precondition, and even there one-past-the-end is allowed. Neither "across" nor "beyond" fixes that underlying issue.

2. Also, documenting this looks to me to be outside the FLS's intended purview, and the FLS has said so before regarding library specifics. The in-bounds requirement is a contract of those library functions (it remains UB to violate it — the question is only where that gets documented), and the changelog has twice recorded the decision not to document exactly this class of rule: rust-lang/rust#124921 ("offset_from: always allow pointers to point to the same address") and rust-lang/rust#117329 ("Allow zero-byte offsets...") are both annotated "This lifted restriction was not documented in the FLS". The language itself has no raw-pointer arithmetic operators to attach the rule to (place projections are specified separately, in the expressions chapter).

The access rule (`fls_c3DaCLQEBpYQ`) already makes any non-zero-sized memory access through an out-of-bounds pointer UB, which is the part the FLS needs to state here. (Zero-sized accesses need no provenance per `core::ptr`, which is another reason to keep the bounds rule out of a blanket UB statement.)

src/values.rst
```diff
@@ -306,3 +324,30 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
...
+
+.. _fls_Dqk4eIvxHloY:
+
+Provenance
+----------
+
+.. rubric:: Legality Rules
+
+:dp:`fls_5MkKtNL9oCsL`
+:t:`Provenance` is an optional property of :t:`[pointer]s` that restricts the addresses a :t:`pointer` may point to, the timespan during which the :t:`pointer` may point to those addresses, and whether the :t:`pointer` can read from and write to those addresses.
```
PLeVasseur: In the case of this text "restricts the addresses a pointer may point to", there may be some imprecision: provenance itself doesn't restrict the stored address. A pointer may hold any address (that's precisely
why out-of-bounds `wrapping_add` is legal). Individual operations impose their own validity requirements, but what *provenance* restricts is access. RFC 3559: "The pointer's provenance says where and when the pointer is allowed to access memory."

```suggestion
:t:`Provenance` is an optional property of :t:`[pointer]s` that restricts the
memory locations the :t:`pointer` may access, the timespan during which the
accesses may occur, and whether the accesses may read from or write to memory.
```

I like this updated triad (spatial/temporal/mutability) since it now matches `core::ptr`, stated about access, and "memory locations" matches the wording the PR now uses for :t:`pointer type` ("values indicate memory locations").

src/values.rst
```diff
@@ -306,3 +324,30 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
...
+:dp:`fls_wnJmQYT7iKQf`
+A :t:`derived pointer` inherits the :t:`provenance` of the :t:`pointer` it was created from, if any.
+
+:dp:`fls_ffh8mAkebORJ`
+A :t:`well-formed pointer` is a :t:`pointer` with :t:`provenance`, where all bytes that comprise the :t:`pointer` are initialized, correctly ordered, and are fragments of the same :t:`original pointer`.
+
+.. rubric:: Undefined Behavior
+
+:dp:`fls_c3DaCLQEBpYQ`
+It is undefined behavior to access memory through a :t:`pointer` which does not have :t:`provenance` over that memory.
```
PLeVasseur: In this case, `fls_c3DaCLQEBpYQ` ("does not have provenance over that memory") seems to miss the mutability axis: a write through a shared reference's provenance targets memory the pointer *does* have provenance over, just not for writing.

Proposed:

```suggestion
It is undefined behavior to access memory through a :t:`pointer` that does not
have :t:`provenance` permitting the access.
```

src/values.rst
```diff
@@ -188,6 +200,12 @@ A :t:`static` shall have a :t:`static initializer`, unless it is an
 The :t:`expression` of a :t:`static initializer` shall be a
 :t:`constant expression`.
 
+:dp:`fls_41zoQ54MiLMm`
+The :t:`static` is determined by evaluating its :t:`static initializer`.
```
PLeVasseur: The core issue to me looks to be duplication: `fls_w0nb0mphho7b` already states "The elaboration of a static evaluates its static initializer", and this paragraph exists appears to only exist to set up `fls_37oocZVDne5Y`, which reads fine on its own to me.

Perhaps we can delete this paragraph outright?

That also dissolves the "value of a value" expansion problem that motivated the earlier suggestion. The analogous constant paragraph `fls_deuo1pn8cjd6` is fine to keep as-is: it predates this PR, and `constant` is glossary-defined as an expression whose uses are substituted by the value, so "the value of a constant" doesn't hit the same expansion.

src/values.rst
```diff
@@ -42,6 +42,15 @@ Two :t:`[value]s` :t:`overlap` when
 * :dp:`fls_eoak5mdl6ma`
   Both :t:`[value]s` are elements of the same :t:`array`.
 
+:dp:`fls_jriT46yWgIR0`
+A :t:`pointer` is a :t:`value` of a :t:`pointer type`.
+
+:dp:`fls_VWUlxTy0QF9d`
+An :t:`original pointer` is a :t:`pointer` created via allocation.
```
PLeVasseur: Thinking about it, it seems that both definitions might be narrower than the behavior they need to cover.

"created via allocation": the common originals involve no allocation call. `&x` and `&raw const x` produce provenance-carrying pointers from an allocated object that already exists — and per the FLS's own glossary a raw borrow expression doesn't even incur a borrow, so neither "allocation" nor "borrow" describes that case. Since the FLS now has `` :t:`allocated object` `` (`fls_CUJyMj0Sj8NS`), the split seems to work is by *origin*: created directly from an allocated object, vs obtained from another pointer.

"address or pointer arithmetic" under-enumerates derivation: casts, copies, and loads of stored pointers from memory also propagate provenance (the [`core::ptr` provenance section](https://doc.rust-lang.org/core/ptr/index.html#provenance) walks through the provenance-preserving operations) — a loaded pointer is in effect a copy of the stored one, so the inheritance rule `fls_wnJmQYT7iKQf` (derived pointers inherit provenance) should apply to it. I'd drop the enumeration entirely — "obtained from another :t:`pointer`" — so the definition can't be incomplete.

src/values.rst
```diff
@@ -306,3 +324,30 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
...
+Provenance
+----------
+
+.. rubric:: Legality Rules
+
+:dp:`fls_5MkKtNL9oCsL`
+:t:`Provenance` is an optional property of :t:`[pointer]s` that restricts the addresses a :t:`pointer` may point to, the timespan during which the :t:`pointer` may point to those addresses, and whether the :t:`pointer` can read from and write to those addresses.
+
+:dp:`fls_1NJhTBN1D2qv`
+An :t:`original pointer` always carries :t:`provenance`.
```
PLeVasseur: Also might be worth tightening `fls_1NJhTBN1D2qv` from "always carries provenance" to say over what:

```suggestion
An :t:`original pointer` carries :t:`provenance` over all or part of the
:t:`allocated object` it was created from.
```

"All or part" because `core::ptr` only commits to "at least the range that slice describes" for references and notes that whether provenance gets shrunk to fit is not yet determined — the FLS shouldn't promise more extent than upstream does.

src/types-and-traits.rst
```diff
@@ -939,6 +939,11 @@ A :t:`function item type` implements the :std:`core::ops::Fn` :t:`trait`, the :s
 Indirection Types
 -----------------
 
+.. rubric:: Legality Rules
+
+:dp:`fls_3qI8FXMsyk0f`
+A :t:`pointer type` is a :t:`type` whose :t:`[value]s` indicate memory locations.
```
PLeVasseur: Reading through here I think I spotted one scope problem and one naming collision in the new `pointer type` definition, both fixable with the same edit:

1. **Scope.** "A type whose values indicate memory locations" describes function pointer types too — a function pointer's value indicates where code lives, which is exactly why `indirection type` ("a type whose values refer to memory locations") includes them. So as defined, function pointers are `pointer`s, and the provenance rules, the well-formed pointer requirement on const/static values, and the access/offset UB rules all textually apply to them. But whether function pointers have provenance at all is an open question upstream — [rust-lang/unsafe-code-guidelines#340](https://github.com/rust-lang/unsafe-code-guidelines/issues/340), the issue bjorn3 pointed to on Zulip ("References have provenance. I believe it hasn't been decided yet if function pointers have provenance"). The FLS shouldn't answer that question as a side effect of a definition's wording.

2. **Duplication.** `pointer type` and the pre-existing `indirection type` now coexist with definitions that don't distinguish them ("indicate memory locations" vs "refer to memory locations"). Two terms, one meaning, in the same section.

Defining `pointer type` by enumeration fixes both at once, and scopes the term to exactly the two cases that have been decided — references have provenance, and raw pointers uncontroversially do:

```suggestion
A :t:`pointer type` is either a :t:`raw pointer type` or a :t:`reference
type`.
```

(same wording with `:dt:` for the glossary entry).

That keeps the provenance rules from taking a position UCG#340 hasn't, and gives `pointer type` a
meaning `indirection type` doesn't already have. It does leave function-pointer values in const/static final values unconstrained by the FLS — deliberate underspecification until UCG#340 resolves, and worth one line on the record (maybe as an issue we open?) so the gap reads as chosen rather than missed.

src/values.rst
```diff
@@ -306,3 +324,30 @@ promoting the produced :t:`mutable borrow` to have ``'static`` :t:`lifetime`.
 #. :dp:`fls_ap85svxyuhvg`
    The :t:`value` of the anonymous :t:`constant` is :t:`borrowed` with
    ``'static`` :t:`lifetime`.
+
+.. _fls_Dqk4eIvxHloY:
+
+Provenance
```
PLeVasseur: I had some readability rather than correctness thoughts on this part.

While forward references are legal in the FLS, this PR splits the material three ways: the definitions of pointer / original pointer / derived pointer sit in the chapter intro, the provenance rules that give them meaning live in a section after Constant Promotion, and the Constants/Statics legality rules reference `well-formed pointer`, defined at the end of the chapter. 

Perhaps we can consolidate all the provenance definitions and rules into the Provenance section and then place that section immediately after the chapter intro, before Constants. Everything then reads top-down, and it keeps the intro focused on the value / allocated-object / overlap fundamentals.

src/glossary.rst
PLeVasseur: The update to pointers makes sense along the lines of what Ralf suggested, cool.

I see two possible gaps remaining in the definition (details and rustc examples on the `src/values.rst` thread for `fls_ffh8mAkebORJ`):

1. It still requires provenance, so provenance-free pointer constants that rustc accepts (`core::ptr::null()`, `5 as *const u8`) are still ruled out. The fragment restriction only concerns bytes that carry provenance.
2. "Fragments of the same :t:`original pointer`" over-constrains the source: fragments of a *derived* pointer (e.g. a cast) reassemble fine; the compiler only needs the bytes to rebuild a single pointer value with provenance.

For this entry:

```suggestion
A :dt:`well-formed pointer` is a :t:`pointer` where either no byte of the
:t:`pointer` carries :t:`provenance`, or every byte of the :t:`pointer` is the
corresponding byte of a single :t:`pointer` with :t:`provenance`.
```

("initialized" is fine to keep or drop — it's already implied by `fls_6lg0oaaopc26`.)

src/values.rst
PLeVasseur: This is not intended to be addressed in #707, but it's something I noticed we could do a follow-up on. If this looks reasonable I can go open an issue for it. Lemme know.

The well-formedness rules being added in #707 cover the pointer-fragment restriction from rust-lang/rust#148259, but rustc enforces a second condition on final values that the FLS won't yet capture: a provenance-carrying pointer in the final value must not point to an allocation freed during evaluation.

```rust
const P: *const u8 = { let x = 0u8; &raw const x };
```

Current rustc [rejects](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fdd9858095fe8b6b205b782283457b2c) this with "encountered dangling pointer in final value of constant". Note that it seems this predates rust-lang/rust#148259.

Scope note: the condition attaches to pointers *in the final value*. Allocations created and freed during evaluation are fine as long as no provenance-carrying pointer to them survives into the final value.

Note the existing FLS `dangling` term can't be reused as-is: it counts null as dangling, while `const N: *const u8 = core::ptr::null();` is a legal constant. A liveness-only condition (the provenance's allocated object still exists at the end of evaluation) would need its own wording.

PLeVasseur (changes requested):
Hey @kirtchev-adacore, thanks for taking the time to write this up.

I spotted some things we could tune up here and there for precision and to match current understandings for provenance that seem to exist in the Project.

I know the release is tomorrow and we meet Friday, but hopefully this is not too much 😅 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/f7b0d8703918dba4cbf11344f7d48119fcd75871..a8393ef30e197c134f32968cd6ae507fbb010f81)*

src/values.rst
kirtchev-adacore: I like it, incorporated!

src/values.rst
kirtchev-adacore: I wan't aware of the compiler behavior, incorporated!

src/values.rst
kirtchev-adacore: Incorporated.

src/values.rst
kirtchev-adacore: Indeed, the FLS wording would need to depend on details from `core`. So I agree, the rule should be removed.

Removed.

src/values.rst
kirtchev-adacore: I like it, incorporated!

src/values.rst
kirtchev-adacore: Incorporated.

src/values.rst
kirtchev-adacore: Removed.

kirtchev-adacore: I also removed an identical paragraph from constants, for the same reasons.

src/values.rst
kirtchev-adacore: Incorporated.

src/types-and-traits.rst
kirtchev-adacore: Incorporated.

src/values.rst
kirtchev-adacore: How about adding "non-dangling" to the definition of "well-formed pointer"?
```rst
A :dt:`well-formed pointer` is a non-dangling :t:`pointer` where either no byte
of the :t:`pointer` carries :t:`provenance`, or every byte of the :t:`pointer`
is the corresponding byte of a single :t:`pointer` with :t:`provenance`.
```

src/values.rst
kirtchev-adacore: I like it! (I actually read the first two paragraphs of your message, then wrote up basically what you said in the third paragraphs :laughing: It pays to read everything!)

src/glossary.rst
kirtchev-adacore: Incorporated.

src/values.rst
kirtchev-adacore: The Provenance section of `core::ptr` explicitly states:

> When an [allocation](https://doc.rust-lang.org/core/ptr/index.html#allocation) is created, it has a unique Original Pointer.

Perhaps an original pointer should be
```rst
An :dt:`original pointer` is a unique :t:`pointer` for an allocation.
```

> &x and &raw const x produce provenance-carrying pointers

Correct, but they are derived pointers - "Provenance is implicitly inherited by all pointers transitively derived from the Original Pointer"

The definition of "derived pointer" can be expanded to
```rst
A :dt:`derived pointer` is a :t:`pointer` obtained by either :t:`borrowing`, copying
a :t:`pointer`, loading a stored :t:`pointer`, performing :t:`pointer` arithmetic,
or performing a :t:`pointer` :t:`cast`.
```

PLeVasseur: Yep, you're right that the `core::ptr` quote settles it; I had the model wrong on borrows. If each allocation has a unique Original Pointer and provenance is "implicitly inherited by all pointers transitively derived" from it, then `&x` and `&raw const x` are derived.

Two refinements we could consider on the wording, both small:

1. For the original-pointer definition, "a unique pointer for an allocation" leaves the uniqueness ambiguous (unique in what sense?), and the FLS already has the term for the memory side. Here's an attempt to tie both down:

   ```suggestion
   An :t:`original pointer` is the unique :t:`pointer` produced when an
   :t:`allocated object` is created.
   ```

   ("the unique pointer produced when ... is created" is the `core::ptr` sentence restated with the FLS's own term; `:dt:` variant for the glossary.)

2. Your expanded derived list looks correct to me! I poked around for provenance-preserving operations that escape it and came up empty: `with_addr`/`map_addr` are documented as equivalent to wrapping pointer arithmetic, union type punning is a load, argument passing is a copy, reborrows are borrows. Maybe just a little role/phrasing polish?

   ```suggestion
   A :t:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`,
   copying a :t:`pointer`, loading a stored :t:`pointer`, performing
   :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
   ```

   (`borrowing` and `casting` look to be existing glossary terms, so maybe we link and use them?)

One optional consequence of your model, maybe to consider: with borrows demoted to derived, the Original Pointer's provenance must cover everything any derived pointer can ever access — i.e. the entire allocated object. So `fls_1NJhTBN1D2qv` could now drop "all or part" and say the original pointer carries provenance over the allocated object it was created from, with all narrowing living in derivation. "All or part" stays true either way, so maybe not a big deal.

src/values.rst
PLeVasseur: Thanks for going for it, but "non-dangling" specifically would reintroduce the exact bug we just fixed, through the side door. The FLS's `dangling` is defined as "either null or not all of the bytes at the referred memory location are part of the same allocation", so a non-dangling requirement in `well-formed pointer` makes these illegal again:

```rust
const N: *const u8 = core::ptr::null();               // null => dangling
const D: *const u8 = core::ptr::without_provenance(16); // no allocation => dangling
```

Both compile today, and both made it through through the definition we just landed via the no-provenance arm.

It looks like the compiler's actual condition is narrower on two axes:

1. It only bites pointers *with provenance*, so null and `without_provenance` pointers are exempt by construction.
2. This is a property of the *final value*, not of pointers generally. Holding a dangling pointer during evaluation [seems fine](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3238b0f8dfc98e4e5205ae9231888991):

   ```rust
    const OK: u8 = {
        let p = { let x = 5u8; &raw const x };
        // x is gone; p dangles for the rest of evaluation — fine,
        // because p never reaches the final value.
        let _ = p;
        0
    };
   ```

   `p` dangles the moment `x` drops, and this compiles.

So folding it into `well-formed pointer` puts a final-value-only condition into a term that reads like a general pointer property which might be a bit like a trap for whoever reuses the term next.

Maybe we can keep #707 as it now stands and do this as the follow-up: some kind of separate legality rule on the const/static side, something in the direction of "if the value ... contains a pointer with provenance, the allocated object the provenance covers shall not have been deallocated during evaluation."

If that seems okay, I can open the issue with that draft and a link back here.

src/values.rst
PLeVasseur: Ha! Convergent evolution is the best review outcome :)

I hear crabs are one possible state: [carcinisation](https://en.wikipedia.org/wiki/Carcinisation)! 🦀

PLeVasseur (dismissed):
Thanks for incorporating the ideas.
Really just a couple of small nits left.

Perhaps we punt on one of the items I raised? I can open an issue.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/f7b0d8703918dba4cbf11344f7d48119fcd75871..e1d4a629ca34dd77f1beef56fd7cde0e468cb842)*

src/values.rst
kirtchev-adacore: > An :t:`original pointer` is the unique :t:`pointer` produced when an :t:`allocated object` is created.

Sounds fine, however the FLS does not define `allocated object`, nor `object` for that matter. The Reference is also a little iffy on what "object" is, most of the time it is bundled as "trait object", which is not what we want.

We may want to define what "allocation" is (promptly stolen from [core::ptr](https://doc.rust-lang.org/core/ptr/index.html#allocation)):
```rst
An :dt:`allocation` is a subset of addressable program memory, characterized by a base
address, size, and a set of addresses.
```

With that, we can how say:
```rst
An :t:`original pointer` is a :t:`pointer` created via :t:`allocation`.
```

src/values.rst
kirtchev-adacore: How about this gymnastic then:
```rst
A :dt:`well-formed pointer` is a :t:`pointer` where either no byte
of the :t:`pointer` carries :t:`provenance`, or every byte of the :t:`pointer`
is the corresponding byte of a single :t:`pointer` with :t:`provenance`.
```
(non-dangling removed)
```rst
If the :t:`value` produced by evaluating a :t:`constant initializer` is a :t:`pointer`,
then the :t:`pointer` shall be a non-dangling :t:`well-formed pointer`.
```
(non-dangling added)

src/glossary.rst
```diff
@@ -3476,6 +3485,11 @@ program entry point
 A :dt:`program entry point` is a :t:`function` that is invoked at the start of
 a Rust program.
 
+provenance
+^^^^^^^^^^
+
+:dt:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the accesses may read from or write to memory.
```
tshepang: - I understand "access" as ability to read from the memory locations, such that a read is implied (unless you find the suggestion too vague)
- "write to memory" does not indicate which memory
```suggestion
:dt:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the :t:`pointer` can write to such memory locations.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/f7b0d8703918dba4cbf11344f7d48119fcd75871..979328b4c2722253a8b61385af3aa69d78d50431)*

kirtchev-adacore: Note that you left out reads! This is important, as provenance dictates **read** and **write** permissions.

src/values.rst
tshepang: That definition of allocation does not sound right... expanded, we have "An original pointer is a pointer created via a subset of addressable program memory". Also, I would expect an allocation to be a verb and not a noun. Also had a look at wasm and ada manuals, and the term is not defined, so maybe it's not so bad to not give it a definition here. Otherwise, how about:

> An allocation is a provision of memory locations.

src/glossary.rst
tshepang: access sounds like it implies reads, at least in the context of a programming language

src/values.rst
kirtchev-adacore: Note that Ralf was very adamant about using terminology and notions that are already established within the Project, and thus within the Rust community at large. `core::ptr` defines "allocation" as

An allocation is a subset of program memory which is addressable from Rust, and within which pointer arithmetic is possible. Examples of allocations include heap allocations, stack-allocated variables, statics, and consts. . . . An allocation has a base address, a size, and a set of memory addresses.

I am inclined to stick with the definition given in `core::ptr`.

Also, the verb is "to allocate", and the noun is "allocation".

src/glossary.rst
kirtchev-adacore: I do not agree. "Memory access" has always meant reads and writes. You have RAM (Random **Access** Memory), which is both read and written, as opposed to ROM (Read-**only** Memory).

Also, from `core::ptr`:

Whether a pointer is valid depends on the operation it is used for (read or write), and the extent of the memory that is **accessed** (i.e., how many bytes are **read/written**)

If you do not like the verb "to access", then you will have to suggest a better alternative.

src/values.rst
tshepang: hm, then were are talking about different things... core::ptr is talking about allocation as a result of some action (what I call "provision"), but your proposed definition talks about it as that action (hence my calling it a verb)

here repeating my expansion of your proposed definition of "original pointer", to make it clear that something needs to change

> An original pointer is a pointer created via a subset of addressable program memory

how about

> An original pointer is a pointer created via provisioning a subset of addressable program memory

then compressed to

> An original pointer is a pointer created via provisioning an allocation

happy to hear a better word than provisioning

we can alternatively not try to define "allocation" (same as those specs I mentioned do), then this problem goes away

kirtchev-adacore: I think we have to define what an "allocation" is because our fallback clause [1.3:3](https://rust-lang.github.io/fls/general.html#definitions) takes us to [allocation](https://www.merriam-webster.com/dictionary/allocation), which does not have a computer science-specific definition. Plus "allocation" is the distinguishing factor between an original pointer and the other kinds of pointers.

Merriam-Webster defines [provision](https://www.merriam-webster.com/dictionary/provision) as ": to supply with needed materials (such as food) : to supply with provisions", but I have never heard of an "allocation being provisioned" in the context of computer science. Allocations are usually created.

I personally think that
```rst
An :t:`original pointer` is a :t:`pointer` created via :t:`allocation`.
```
is fine, and grammatically correct. Here "allocation" is "the act of allocating", which has a verb-like quality. I do not think you can substitute "allocation" with its definition directly because the definition is for the "simple" noun, and here you have a verb-like noun (no idea what these are called in the English grammar). The proper substitution is
```rst
An original pointer is a pointer created by allocating a subset of addressable
program memory, which is characterized by a base address, size, and a set of
addresses.
```

kirtchev-adacore: After a live discussion with Tshepang, we decided not to define what an "allocation" is, and leave the definition of "original pointer" as
```rst
An :t:`original pointer` is a :t:`pointer` created via :t:`allocation`.
```

src/values.rst
kirtchev-adacore: I applied my suggestion.

src/glossary.rst
```diff
@@ -1156,6 +1156,11 @@ A :dt:`derive macro` is a :t:`procedural macro` that consumes a stream of
 :t:`[token]s` and produces a stream of tokens, and is invoked via attribute
 :c:`derive`.
 
+derived pointer
+^^^^^^^^^^^^^^^
+
+A :dt:`derived pointer` is a :t:`pointer` obtained by :t:`borrowing`, copying a :t:`pointer`, loading a stored :t:`pointer`, performing :t:`pointer` arithmetic, or :t:`casting` a :t:`pointer`.
```
tshepang: does "loading" mean reading

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/51b787946f8e3f5cbbf5df632e1c50f31f8d9ed6..bed3e05da1c9acd5db8850be5b60fbaa2a2c1a6b)*

src/values.rst
```diff
@@ -62,6 +62,39 @@ It is undefined behavior to create an :t:`allocated object` with :t:`memory
...
+:t:`Provenance` is an optional property of :t:`[pointer]s` that restricts the memory locations the :t:`pointer` may access, the timespan during which the accesses may occur, and whether the accesses may read from or write to said memory locations.
+
+:dp:`fls_1NJhTBN1D2qv`
+An :t:`original pointer` carries :t:`provenance` over all or part of the :t:`allocated object` it was created from.
+
+:dp:`fls_wnJmQYT7iKQf`
+A :t:`derived pointer` inherits the :t:`provenance` of the :t:`pointer` it was created from, if any.
+
+:dp:`fls_ffh8mAkebORJ`
+A :t:`well-formed pointer` is a :t:`pointer` where either no byte of the :t:`pointer` carries :t:`provenance`, or every byte of the :t:`pointer` is the corresponding byte of a single :t:`pointer` with :t:`provenance`.
```
tshepang, line 91: what does this correspond to... is it in relation to some other pointer, perhaps an original pointer

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/51b787946f8e3f5cbbf5df632e1c50f31f8d9ed6..bed3e05da1c9acd5db8850be5b60fbaa2a2c1a6b)*

src/glossary.rst
kirtchev-adacore: It most likely involves a read of the actual address, but also of the pointer metadata (if any).

src/values.rst
```diff
@@ -62,6 +62,39 @@ It is undefined behavior to create an :t:`allocated object` with :t:`memory
...
+
+.. rubric:: Undefined Behavior
+
+:dp:`fls_c3DaCLQEBpYQ`
+It is undefined behavior to access memory through a :t:`pointer` that does not have :t:`provenance` permitting the access.
```
tshepang, line 96: based on my thinking that provenance implies permission, therefore the word "permission" does not need to be mentioned
```suggestion
It is undefined behavior to access memory through a :t:`pointer` that does not have :t:`provenance` over that memory.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/51b787946f8e3f5cbbf5df632e1c50f31f8d9ed6..bed3e05da1c9acd5db8850be5b60fbaa2a2c1a6b)*

src/glossary.rst
tshepang: let's use "read" instead

src/values.rst
kirtchev-adacore: Let's break this down:
```rst
A :t:`well-formed pointer` is a :t:`pointer` where either no byte of the
:t:`pointer` carries :t:`provenance`
```
A pointer without provenance is a derived pointer that was not obtained from an original pointer. Such derived pointers are created by casting for example.

```rst
A :t:`well-formed pointer` is a :t:`pointer` where ... every byte of the
:t:`pointer` is the corresponding byte of a single :t:`pointer` with :t:`provenance`.
```
An original pointer has provenance by definition. A derived pointer created from an original pointer has provenance by definition.

So there you have it. Both original and derived pointers are involved in this semantic.

src/values.rst
kirtchev-adacore: No, you are not correct here.

I can obtain a mutable raw pointer from a raw pointer by transmutation, but the associated provenance does not grant write access. This is why the rule explicitly states "permitting the access".

src/glossary.rst
tshepang: If access means read and/or write, then we are here saying: "restricts the memory locations the pointer may read-write, when those reads-writes can happen, and whether those are reads and/or writes".

That is, "access" implies reads, but only includes writes optionally (hence my suggestion).

An alternative, in case that still doesn't work (to avoid "access'):

> Provenance is an optional property of pointers that restricts the memory locations the pointer may read, the timespan during which the read may occur, and whether the pointer can also write to such memory locations.

It's a little vague, so maybe:

> Provenance is an optional property of pointers that restricts the memory locations the pointer may read, the timespan during which the read may occur, and whether those reads can also include writes.

Not sure. What do you think.

kirtchev-adacore: More like "and whether the reads and/or writes are allowed in the first place".

I do not think your suggestion is correct, as once again you are leaving out writes from several of the provenance properties. I am not sure why you are trying to "save on text", and cause the rule to imply additional semantics. One of the goals of the FLS is to spell out rules clearly, even if there is repetition, so that there is no doubt, implication, or ambiguity.

Basically "provenance" says
- You are allowed to access memory locations from 0x000A to 0x001A.
- You are allowed to access those memory locations from this part of your program to this part of your program.
- You are allowed to read and/or write from/to those memory locations.

There is "write-only" memory from the point of view of your program! For example, GPU memory can behave this way. Your program simply dumps data onto the GPU buffers, and the GPU reads it. Your program does not read from the GPU.

src/changelog.rst
```diff
@@ -163,6 +163,43 @@ Language changes in Rust 1.95.0
...
 - `const-eval: be more consistent in the behavior of padding during typed copies <https://github.com/rust-lang/rust/pull/148967>`_
+
+  New glossary entries:
+
+  - :t:`derived pointer`
+  - :t:`original pointer`
+  - :t:`provenance`
+  - :t:`well-formed pointer`
+
+  Changed glossary entry: :t:`raw pointer`
```
tshepang: I would not bother adding glossary entries to changelog (don't remember what we decided on), but if you have to insist, I would add "pinter type" here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/707/51b787946f8e3f5cbbf5df632e1c50f31f8d9ed6..5346bdce1352f14917de2a380f97f61ee87e1002)*

kirtchev-adacore: We already have such changelog entries for 1.93.0. We agreed to include such entries, until the glossary is automatically generated.

tshepang (approved):

src/glossary.rst
tshepang: > There is "write-only" memory from the point of view of your program! For example, GPU memory can behave this way.

good to know

PR #708, opened 2026-07-03
Add changelog entry for 1.96.1

kirtchev-adacore:


src/changelog.rst
tshepang, line 1: section title is "Language changes in Rust 1.96.1", and these are not language changes... I suggest we don't add this section, or we just say "none"

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/708/3c73fb2263ce3ffc9f7d7592a910d42372cad001..1a5b76aecd4ed9e2c2da899cfc6e021da518266c)*

kirtchev-adacore: We still want traceability between what went into the Rust release, and why it was not included in the FLS.

kirtchev-adacore:
We agreed that patch releases are not documented in the Changelog of the FLS.

Issue #711, opened 2026-07-09
Use-import prefix construction does not preserve leading `::`


PLeVasseur:
## Problem

The FLS use-import model constructs an `import path prefix` primarily from path segments. A leading `::`, however, is a namespace qualifier rather than a path segment, so the construction in `fls_WAA4WmohGu6T` does not clearly retain it.

The grammar permits a use-import prefix consisting only of `::`, while the related prefix definition and construction describe prefixes in terms of a leading simple path or its path segments. This leaves the FLS unable to rigorously distinguish these Rust 2021 cases:

```rust
use {self as this_module};      // Valid: empty local prefix.
use ::{self as root};           // Invalid: imports the bare extern-prelude root.
use ::std::{self as std_alias}; // Valid: global path through the extern prelude.
```

In particular:

- An empty prefix with no leading `::` should resolve from the current module.
- A bare `::` prefix denotes the extern-prelude root and cannot be imported.
- A prefix such as `::std` must retain `::`; otherwise it can become indistinguishable from locally resolved `std`.
- Prefix construction through nested use trees must retain `::` while ensuring it remains at the start of the fully combined path.

This problem predates #694. That PR exposes it by adding a restriction that refers directly to an import path prefix consisting only of `::`.

## Expected result

The FLS should model the complete use-import prefix, including any leading namespace qualifier, and align the grammar, prefix definitions, construction rules, and legality rules.

The resulting model should account for:

```rust
// Valid.
use {self as this_module};
use ::std::{self as std_alias};
use {::{std::mem as m}};

// Invalid.
use ::{self as root};
use ::*;
use ::{*};
use std::{::{std::mem as m}};
use ::{::{std::mem as m}};
```

This issue is limited to the inherited leading-`::`/import-prefix model. The keyword-import restrictions added by #694 and unrelated `self` wording cleanups do not need to be reopened here.

## References

- Follow-up discussion from #694: https://github.com/rust-lang/fls/pull/694#discussion_r3547319628
- Rust Reference, use-declaration restrictions: https://doc.rust-lang.org/reference/items/use-declarations.html#r-items.use.restrictions.extern-prelude
- Rust Reference, brace syntax: https://doc.rust-lang.org/reference/items/use-declarations.html#r-items.use.multiple-syntax

PR #712, opened 2026-07-14
[Dev-Guide]: Dealing with PR blockers

kirtchev-adacore:


src/dev-guide.rst
tshepang, line 1: not sure we should limit discussion to team meets... I think it's fine to just say "bring up the issue to the team"

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/712/3977e597187d55140d57c75227a116b6afc7a852..c69778425b5c44c215fb2cbc7c5dc4df697ecfa3)*

kirtchev-adacore: Works for me! Updated.

tshepang (approved):
approving, but maybe better to wait for @PLeVasseur approval also

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/712/3977e597187d55140d57c75227a116b6afc7a852..d90582d0716d749d22cdcc31588882409a58cee6)*

src/dev-guide.rst
```diff
@@ -96,6 +96,30 @@ Use the following sentence pattern for multiple syntax categories::
...
+
+Incorrect or incomplete semantics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While working on a PR, a developer may need to modify existing FLS semantics or terminology, only for the developer to discover that these semantics or terms are either incorrect or incomplete.
+In such cases, bring up the issue to the FLS team.
+
+Depending on the scope of the required changes, the FLS team may decide to either:
+- Enact the necessary changes in the same PR-in-development, or
+- Approximate the contents of the PR-in-development, and open a GitHub issue to track the update on the semantics or terms.
```
PLeVasseur: Could we state the condition under which deferring the correction is acceptable? Like we chatted about in our meeting,  in #694 the leading-`::` modeling defect predated the PR and we could keep that PR narrowly scoped without making the defect worse, then track the inherited problem in #711.

As written, "approximate" also appears to permit merging an error introduced by the current PR. I think this option should instead say that a pre-existing, out-of-scope problem may be deferred only when the current PR remains accurate within its stated scope. Otherwise the correction needs to land in the under-review PR or in a prerequisite PR.

I'm a little worried about this being too loose for the conditions under which we can punt to create an issue and address later. Hope you catch my drift.

src/dev-guide.rst
```diff
@@ -96,6 +96,30 @@ Use the following sentence pattern for multiple syntax categories::
...
+
+Missing semantics
+~~~~~~~~~~~~~~~~~
+
+While working on a PR, a developer may need to modify existing FLS semantics or use existing terminology, only for the developer to discover that these semantics or terms are missing.
```
PLeVasseur: This was kinda hard to parse, as I couldn't tell how to apply it. Seems this branch details the semantics are described as both existing in the FLS and missing from it, sorta.

Could we state the trigger a little more bluntly / directly? For example:

```suggestion
While working on a PR, a developer may discover that the PR depends on semantics or terminology that the FLS does not yet define.
```

This would also distinguish this case from incomplete existing text in the preceding subsection.

PLeVasseur (changes requested):
Thanks for taking this up @kirtchev-adacore!

I had some thoughts on where we could tighten things up a bit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/712/b26f6f6a0b0594afba7fccabc61234ebec212bfb..3335b235780ed28da8b8249c583c2c4dd3eced70)*

src/dev-guide.rst
kirtchev-adacore: Incorporated.

src/dev-guide.rst
kirtchev-adacore: Reworded.

PLeVasseur (approved):
LGTM. Thank you for adding this, @kirtchev-adacore!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/712/b26f6f6a0b0594afba7fccabc61234ebec212bfb..007d2c5c6a0e7899cbff343c20eebf485502e289)*

PR #715, opened 2026-07-28
Remove term "indirection" and its derivatives

### LINKED ISSUE #710: [Change]: Remove the term "indirection", and associated derivatives
### Change description

Remove the terms "indirection type", "indirection value", and similar. The appropriate parallel terms are "pointer type", "pointer", etc.

### Area

N/A

### Category

Bug

### FLS Paragraph ID

_No response_

### Change description (extended)

Remove the terms "indirection type", "indirection value", and similar. The appropriate parallel terms are "pointer type", "pointer", etc.

kirtchev-adacore:
This PR removes the term "indirection" and its derivatives from the FLS as the Rust Project does not use this term. The preferred alternative is "pointer", "pointer type", etc.

Closes: https://github.com/rust-lang/fls/issues/710

### DIFF
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -22,7 +22,7 @@ with the change that has been applied due to it.
 FLS maintenance
 ---------------
 
-- `[Change]: Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
+- `Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
 
   Changed glossary entries:
 
@@ -43,19 +43,9 @@ FLS maintenance
   - :p:`fls_jrohsv7hx7yw`
   - :p:`fls_3qI8FXMsyk0f`
   - :p:`fls_rpbhr0xukbx9`
-  - :p:`fls_8uWfFAsZeRCs`
   - :p:`fls_twhq24s8kchh`
-  - :p:`fls_w4NbA7WhZfR2`
-  - :p:`fls_ie0avzljmxfm`
   - :p:`fls_15zdiqsm1q3p`
-  - :p:`fls_GUZuiST7ucib`
-  - :p:`fls_vaas9kns4zo6`
-  - :p:`fls_n6ffcms5pr0r`
-  - :p:`fls_26Xgem831Nqg`
   - :p:`fls_ozYgHEHFTT5c`
-  - :p:`fls_18ke90udyp67`
-  - :p:`fls_nrqG8i3fmpm4`
-  - :p:`fls_e5hivr6m5s3h`
 
   Moved paragraphs:
 
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -22,7 +22,7 @@ with the change that has been applied due to it.
 FLS maintenance
 ---------------
 
-- `Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
+- Remove the term "indirection", and associated derivatives
 
   Changed glossary entries:
 
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -19,6 +19,13 @@ with the change that has been applied due to it.
    just the language changes that had an impact to the FLS. See the `release
    notes`_ for a full list of changes.
 
+FLS maintenance
+---------------
+
+- `[Change]: Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
+
+  TODO: Update when things settle.
+
 Language changes in Rust 1.98.0
 -------------------------------
 
@@ -119,7 +126,7 @@ Language changes in Rust 1.95.0
 
   - :t:`let binding`
 
-  Changed existing paragraphs:
+  Changed paragraphs:
 
   - :p:`fls_72JHo343O7jp`
   - :p:`fls_6bwTtGKb7ba7`
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -24,7 +24,84 @@ FLS maintenance
 
 - `[Change]: Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
 
-  TODO: Update when things settle.
+  Changed glossary entries:
+
+  - :t:`dangling`
+  - :t:`fat pointer type`
+  - :t:`function pointer type`
+  - :t:`pointer type`
+  - :t:`raw pointer type`
+  - :t:`reference type`
+  - :t:`thin pointer type`
+
+  Removed glossary entries:
+
+  - indirection type
+
+  Changed paragraphs:
+
+  - :p:`fls_jrohsv7hx7yw`
+  - :p:`fls_3qI8FXMsyk0f`
+  - :p:`fls_rpbhr0xukbx9`
+  - :p:`fls_8uWfFAsZeRCs`
+  - :p:`fls_twhq24s8kchh`
+  - :p:`fls_w4NbA7WhZfR2`
+  - :p:`fls_ie0avzljmxfm`
+  - :p:`fls_15zdiqsm1q3p`
+  - :p:`fls_GUZuiST7ucib`
+  - :p:`fls_vaas9kns4zo6`
+  - :p:`fls_n6ffcms5pr0r`
+  - :p:`fls_26Xgem831Nqg`
+  - :p:`fls_ozYgHEHFTT5c`
+  - :p:`fls_18ke90udyp67`
+  - :p:`fls_nrqG8i3fmpm4`
+  - :p:`fls_e5hivr6m5s3h`
+
+  Moved paragraphs:
+
+  - :p:`fls_1kg1mknf4yx7`
+  - :p:`fls_v2wrytr3t04h`
+  - :p:`fls_5dd7icjcl3nt`
+  - :p:`fls_B0SMXRqQMS1E`
+  - :p:`fls_hbn1l42xmr3h`
+  - :p:`fls_g1iYVw7upBnH`
+  - :p:`fls_8gpvNJfVlyaD`
+  - :p:`fls_KcI6yK0P8Onn`
+  - :p:`fls_52thmi9hnoks`
+  - :p:`fls_bYWfGDAQcWfA`
+  - :p:`fls_c2Guy3fPYaUV`
+  - :p:`fls_hrum767l6dte`
+  - :p:`fls_k6ues2936pjq`
+  - :p:`fls_csdjfwczlzfd`
+  - :p:`fls_ezh8aq6fmdvz`
+  - :p:`fls_jriT46yWgIR0`
+  - :p:`fls_VWUlxTy0QF9d`
+  - :p:`fls_kaPNJ7iIHPro`
+  - :p:`fls_5MkKtNL9oCsL`
+  - :p:`fls_1NJhTBN1D2qv`
+  - :p:`fls_wnJmQYT7iKQf`
+  - :p:`fls_ffh8mAkebORJ`
+  - :p:`fls_c3DaCLQEBpYQ`
+
+  New paragraphs:
+
+  - :p:`fls_Im7miUSS87xs`
+  - :p:`fls_aQgOFrzAhdsC`
+
+  Changed sections:
+
+  - :ref:`fls_3i4ou0dq64ny`
+
+  Moved sections:
+
+  - :ref:`fls_xztr1kebz8bo`
+  - :ref:`fls_Dqk4eIvxHloY`
+
+  New sections:
+
+  - :ref:`fls_xOuhiItK1hK0`
+  - :ref:`fls_qPWQnNXH52R3`
+  - :ref:`fls_phgP8YIwpXpi`
 
 Language changes in Rust 1.98.0
 -------------------------------
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -19,6 +19,44 @@ with the change that has been applied due to it.
    just the language changes that had an impact to the FLS. See the `release
    notes`_ for a full list of changes.
 
+FLS maintenance
+---------------
+
+- `[Change]: Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
+
+  Changed glossary entries:
+
+  - :t:`dangling`
+  - :t:`fat pointer type`
+  - :t:`function pointer type`
+  - :t:`pointer type`
+  - :t:`raw pointer type`
+  - :t:`reference type`
+  - :t:`thin pointer type`
+
+  Removed glossary entries:
+
+  - :t:`indirection type`
+
+  New paragraphs:
+
+  - :p:`fls_f3TcmFrVoNis`
+  - :p:`fls_8RBNIR0E6pnI`
+  - :p:`fls_Im7miUSS87xs`
+  - :p:`fls_aQgOFrzAhdsC`
+
+  Changed paragraphs:
+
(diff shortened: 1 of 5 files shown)


src/types-and-traits.rst
```diff
@@ -936,13 +936,19 @@ A :t:`function item type` implements the :std:`core::ops::Fn` :t:`trait`, the :s
...
 
-Indirection Types
+Pointer Types
 -----------------
 
 .. rubric:: Legality Rules
 
 :dp:`fls_3qI8FXMsyk0f`
-A :t:`pointer type` is either a :t:`raw pointer type` or a :t:`reference type`.
+A :t:`pointer type` is a :t:`type` whose :t:`[value]s` refer to memory locations.
```
PLeVasseur, line 945: I took a look at the [Rust 1.96 Reference classification](https://doc.rust-lang.org/1.96.0/reference/types.html#pointer-types). It classifies function pointers as pointer types, but I don't think this PR propagates that broader meaning to existing FLS rules that use `pointer type` in the narrower, data-pointer sense.

In particular, the existing [FLS dereference rule](https://github.com/rust-lang/fls/blob/5f464ca69852d727230973031957eeae18bf0bf3/src/expressions.rst#L1080-L1113) seems to now apply to function pointers and says that `*fp` denotes a pointed-to memory location, while its type rule falls back to `Deref::Target`.

On the other hand, the Reference's [dereference-operator rule](https://doc.rust-lang.org/1.96.0/reference/expressions/operator-expr.html#the-dereference-operator) uses built-in dereference for data pointers/`Box` and otherwise requires `Deref`, while the [documented function-pointer trait implementations](https://doc.rust-lang.org/1.96.0/std/primitive.fn.html#trait-implementations-1) do not include `Deref`.

Maybe we can define this category extensionally as function pointer, raw pointer, or reference types, then audit and narrow the existing data-pointer rules to raw pointers/references?

It seems to me that the memory-location predicate also might not characterize function pointers, which the Reference defines as [referring to functions](https://doc.rust-lang.org/1.96.0/reference/types/function-pointer.html#function-pointer-types), and does not clearly account for valid null raw-pointer values such as [`core::ptr::null()`](https://doc.rust-lang.org/1.96.0/core/ptr/fn.null.html).

src/types-and-traits.rst
```diff
@@ -936,13 +936,19 @@ A :t:`function item type` implements the :std:`core::ops::Fn` :t:`trait`, the :s
...
+
+:dp:`fls_f3TcmFrVoNis`
+A :t:`pointer` is a :t:`value` of a :t:`pointer type`.
+
+:dp:`fls_8RBNIR0E6pnI`
+A :t:`pointer` is :t:`dangling` if it is either :c:`null` or not all of the bytes at the referred memory location are part of the same allocation.
```
PLeVasseur: Over in the Reference it derives a pointer's [pointed-to bytes](https://doc.rust-lang.org/1.96.0/reference/behavior-considered-undefined.html#pointed-to-bytes) from its value and pointee size, then defines [dangling pointers](https://doc.rust-lang.org/1.96.0/reference/behavior-considered-undefined.html#dangling-pointers) by whether all those bytes belong to one **live** allocation.

It explicitly says a zero-sized pointee is never dangling even when the pointer is null, so `core::ptr::null::<()>()` is a counterexample to this definition. Omitting "live" also leaves a pointer to deallocated storage not clearly covered.

Might be better here if we follow the Reference wording and limit this data-pointer rule to raw pointers/references, since `pointer` now also includes function pointers. What do you think?

PLeVasseur (changes requested):
Thanks @kirtchev-adacore, for taking the time to adopt the Rust Project terminology.

I found a couple of places where we seem to be a bit at odds with what's written up in the Reference. Could you take a look?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/b26f6f6a0b0594afba7fccabc61234ebec212bfb..5f464ca69852d727230973031957eeae18bf0bf3)*

src/types-and-traits.rst
kirtchev-adacore: Indeed, it seems we need a "data pointer type" term for raw pointers and references. So the hierarchy would become
```
pointer type
|
+-- data pointer type
|    |
|    +-- raw pointer type
|    +-- reference type
|
+-- function pointer type
```
Given this hierarchy, all pointer-related semantics would need to be reexamined and recalibrated.

src/types-and-traits.rst
kirtchev-adacore: Yeah, we should probably steal the Reference definition.

Have a look at my analysis of https://github.com/rust-lang/fls/issues/716. We may actually want to involve provenance in the notion of "dangling".

src/types-and-traits.rst
kirtchev-adacore: @PLeVasseur Hmm, now that I took a look at the Reference, function pointer types are separate from (data) pointer types!

Given what you outlined above, it might make more sense to amend our classification to
```
pointer type
|
+-- raw pointer type
+-- reference

function pointer type
```

kirtchev-adacore: I looked at every occurrence of "pointer" in the FLS, and I can say that
- Every rule which involves a raw pointer or a reference either uses those two terms, or "pointer".
- Every rule which involves a raw pointer type or a reference type either uses those two terms, or "pointer type".
- There are separate rules for function pointer types and function pointers.

So it seems that the section "Indirection Types" should be renamed to "Pointer Types", and we should have a **separate** section "Function Pointer Types", similar to what the Reference has.

src/types-and-traits.rst
```diff
@@ -1064,47 +989,92 @@ Reference Types
...
-A :t:`mutable reference type` allows the direct mutation of a referenced
-:t:`value`.
+A :t:`mutable reference type` allows the direct mutation of a referenced :t:`value`.
 
 :dp:`fls_n6ffcms5pr0r`
-A :t:`mutable reference type` does not implement the :std:`copy::marker::Copy`
-:t:`trait`.
+A :t:`mutable reference type` does not implement the :std:`copy::marker::Copy` :t:`trait`.
+
+.. rubric:: Examples
```
tshepang, line 1012: let us remove this (here or in another pr, since it was already existing before the pr) as it repeats what the Syntax section already says without adding anything more

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/680fbc75ec03130b8616a0ebf9dd732b989d4460..69176c1d4ae236bbbca43238f01c63aa3dae6a93)*

src/types-and-traits.rst
```diff
@@ -1405,11 +1375,17 @@ Type Layout
 :dp:`fls_kdbq02iguzgl`
 All :t:`[value]s` have an :t:`alignment` and a :t:`size`.
 
+:dp:`fls_Im7miUSS87xs`
+A :t:`fixed sized type` is a :t:`type` that implements the :std:`core::marker::Sized` :t:`trait`.
+
+:dp:`fls_aQgOFrzAhdsC`
+A :t:`thin pointer type` is a :t:`pointer type` that refers to a :t:`fixed sized type`.
```
tshepang, line 1382: I find it not worth introducing a new term for this
```suggestion
:dp:`fls_aQgOFrzAhdsC`
A :t:`thin pointer type` is a :t:`pointer type` that implements the :std:`core::marker::Sized` :t:`trait`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/680fbc75ec03130b8616a0ebf9dd732b989d4460..69176c1d4ae236bbbca43238f01c63aa3dae6a93)*

src/types-and-traits.rst
tshepang: I see now that this is consistent with the other sections in chapter... ignore me

kirtchev-adacore: OK.

src/types-and-traits.rst
kirtchev-adacore: Except that "fixed sized type" is used in other places in the FLS.

tshepang (approved):

src/types-and-traits.rst
PLeVasseur: The updated hierarchy matches the direction we agreed on in the FLS meeting. This addresses my concern about data-pointer rules leaking onto function pointers. Thanks!

src/types-and-traits.rst
PLeVasseur: Per the [2026-08-28 FLS team meeting](https://hackmd.io/@rust-fls-team/Hkmap5nUMl), we agreed to keep this PR focused on pointer terminology and type distinctions, then handle dangling and pointer UB corrections in issue #716 and a second PR. With that scope, this concern no longer blocks #715.

src/values.rst
```diff
@@ -158,6 +125,95 @@ the :t:`constant`.
...
    const ZERO: u32 = 0;
 
+.. _fls_xOuhiItK1hK0:
+
+Pointers
+--------
+
+.. rubric:: Legality Rules
+
+:dp:`fls_f3TcmFrVoNis`
```
PLeVasseur: This unchanged paragraph moved from `Provenance`, but its ID changed from `fls_jriT46yWgIR0` to `fls_f3TcmFrVoNis`. I was pretty sure that paragraph IDs are supposed to remain stable across FLS revisions, and the Rust 1.95 changelog already seems to already expose the old ID as the public link.

Maybe we retain `fls_jriT46yWgIR0` here and record it as a moved paragraph in the maintenance changelog entry for this PR?

src/changelog.rst
```diff
@@ -19,6 +19,13 @@ with the change that has been applied due to it.
    just the language changes that had an impact to the FLS. See the `release
    notes`_ for a full list of changes.
 
+FLS maintenance
+---------------
+
+- `[Change]: Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
+
+  TODO: Update when things settle.
```
PLeVasseur: Pretty sure this is intended to allow a single update when all the work's done, but wanted to make sure to point it out.

PLeVasseur (dismissed):
This looks like it took a great direction. Thanks @kirtchev-adacore!

There were a couple of procedural things I think need be taken care of, but otherwise we should merge:
* finalize the changelog updates
* maintaining the FLS ID whens shifting from one place in the FLS to another

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/680fbc75ec03130b8616a0ebf9dd732b989d4460..69176c1d4ae236bbbca43238f01c63aa3dae6a93)*

src/changelog.rst
kirtchev-adacore: Yep, I was waiting on the discussion to settle before recording the changes.

src/types-and-traits.rst
kirtchev-adacore: Sweet!

src/values.rst
kirtchev-adacore: This was a mistake on my part!

Good thing you caught it, FLS ids should indeed be stable.

kirtchev-adacore: Fixed.

src/changelog.rst
kirtchev-adacore: Updated.

src/changelog.rst
```diff
@@ -19,6 +19,90 @@ with the change that has been applied due to it.
    just the language changes that had an impact to the FLS. See the `release
    notes`_ for a full list of changes.
 
+FLS maintenance
+---------------
+
+- `[Change]: Remove the term "indirection", and associated derivatives <https://github.com/rust-lang/fls/issues/710>`_
```
tshepang: let's keep it clean
```suggestion
- Remove the term "indirection", and associated derivatives
```

src/changelog.rst
```diff
@@ -19,6 +19,90 @@ with the change that has been applied due to it.
...
+  Removed glossary entries:
+
+  - indirection type
+
+  Changed paragraphs:
+
+  - :p:`fls_jrohsv7hx7yw`
+  - :p:`fls_3qI8FXMsyk0f`
+  - :p:`fls_rpbhr0xukbx9`
+  - :p:`fls_8uWfFAsZeRCs`
```
tshepang: was just a formatting change
```suggestion
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/680fbc75ec03130b8616a0ebf9dd732b989d4460..b3fe5dbce009dfb0d0bd063cfb799a09182685db)*

src/changelog.rst
```diff
@@ -19,6 +19,90 @@ with the change that has been applied due to it.
...
+  - :p:`fls_twhq24s8kchh`
+  - :p:`fls_w4NbA7WhZfR2`
+  - :p:`fls_ie0avzljmxfm`
```
tshepang: same
```suggestion
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/680fbc75ec03130b8616a0ebf9dd732b989d4460..b3fe5dbce009dfb0d0bd063cfb799a09182685db)*

src/changelog.rst
```diff
@@ -19,6 +19,90 @@ with the change that has been applied due to it.
...
+  - :p:`fls_15zdiqsm1q3p`
+  - :p:`fls_GUZuiST7ucib`
+  - :p:`fls_vaas9kns4zo6`
+  - :p:`fls_n6ffcms5pr0r`
+  - :p:`fls_26Xgem831Nqg`
```
tshepang: same
```suggestion
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/680fbc75ec03130b8616a0ebf9dd732b989d4460..b3fe5dbce009dfb0d0bd063cfb799a09182685db)*

src/changelog.rst
```diff
@@ -19,6 +19,90 @@ with the change that has been applied due to it.
...
+  - :p:`fls_ozYgHEHFTT5c`
+  - :p:`fls_18ke90udyp67`
+  - :p:`fls_nrqG8i3fmpm4`
+  - :p:`fls_e5hivr6m5s3h`
```
tshepang: same
```suggestion
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/715/680fbc75ec03130b8616a0ebf9dd732b989d4460..b3fe5dbce009dfb0d0bd063cfb799a09182685db)*

tshepang (changes requested):

src/changelog.rst
kirtchev-adacore: OK for removing `[Change]:`, but I think it is useful to also link to the issue, in case people want to look at any discussions.

src/changelog.rst
kirtchev-adacore: Fixed.

src/changelog.rst
kirtchev-adacore: Fixed.

src/changelog.rst
kirtchev-adacore: Fixed.

src/changelog.rst
kirtchev-adacore: Fixed.

src/changelog.rst
tshepang: they could look at this pr for that context... there is no precedent for this, and I would rather we set rules for it before we practice

kirtchev-adacore: OK, works for me!

tshepang (approved):

PR #717, opened 2026-08-07
Fix leading qualifiers in use-import prefixes

### LINKED ISSUE #711: Use-import prefix construction does not preserve leading `::`

PLeVasseur:
## Summary

Preserve leading `::` in import path prefixes, distinguishing an empty prefix from bare `::` and qualified paths such as `::std`.

Replace the stateful construction with a declarative definition and reject misplaced root qualifiers, including in empty nested groups.

Clarify the construction cases, separator, nesting, and self-specific legality wording in response to review.

Reconcile the shadowing rules with explicit global lookup: local names do not prevent `::std` from selecting the external crate. The broader nonglobal ambiguity model and trait-parent rules are unchanged.

Closes rust-lang/fls#711

## Reference alignment

[Rust Reference: use declarations](https://doc.rust-lang.org/reference/items/use-declarations.html)

[Global paths](https://doc.rust-lang.org/reference/paths.html#r-paths.qualifiers.global-root.edition2018) and [prelude shadowing](https://doc.rust-lang.org/reference/names/scopes.html#r-names.scopes.items.shadow-prelude), for Rust 2021.

## Testing

[CI passes](https://github.com/rust-lang/fls/actions/runs/35748301920).

### DIFF
--- a/src/entities-and-resolution.rst
+++ b/src/entities-and-resolution.rst
@@ -1103,8 +1103,7 @@ A :t:`nesting import` is a :t:`use import` that provides a common
 :t:`common path prefix` for its nested :t:`[use import]s`.
 
 :dp:`fls_iNUBX5fJAI1N`
-A :t:`glob import` outside of a :t:`nesting import` without a :t:`common path
-prefix` is rejected, but may still be consumed by :t:`[macro]s`.
+A :t:`glob import` whose :t:`import path prefix` is empty or consists only of :t:`namespace qualifier` ``::`` is rejected, but may still be consumed by :t:`[macro]s`.
 
 :dp:`fls_wB3fVglLOqbZ`
 It is a static error if two :t:`[glob import]s` import the same :t:`name` in the
@@ -1125,7 +1124,7 @@ A :t:`use import` with a single :t:`path segment` expressed as either :t:`keywor
 When a :t:`path segment` expressed as :t:`keyword` ``super`` is used to import a parent :t:`module`, the imported :t:`entity` shall be subject to a :t:`renaming`.
 
 :dp:`fls_aam34hsRmKU2`
-The :t:`import path prefix` of a :t:`glob import` shall neither be empty nor consist only of :t:`namespace qualifier` ``::``. An :t:`import path prefix` that selects the :t:`entity` of a :t:`simple import` shall not consist only of :t:`namespace qualifier` ``::``.
+An :t:`import path prefix` that selects the :t:`entity` of a :t:`simple import` shall not consist only of :t:`namespace qualifier` ``::``.
 
 :dp:`fls_LV94x3HlpBWk`
 A :t:`simple import` shall not refer to :t:`[enum variant]s` through a :t:`type alias`.
--- a/src/entities-and-resolution.rst
+++ b/src/entities-and-resolution.rst
@@ -1065,7 +1065,10 @@ exported by the :t:`module` or :t:`enum` its :t:`import path prefix` resolves to
 into :t:`scope`.
 
 :dp:`fls_BMtRtjJ7gBKT`
-Each :t:`namespace qualifier` of an :t:`import path prefix` shall either be the first element of the :t:`import path prefix` or appear between two :t:`[path segment]s`.
+A :t:`simple import`, :t:`glob import`, or :t:`nesting import` expressed with a
+leading :t:`namespace qualifier` ``::`` shall not be nested, directly or
+indirectly, within any :t:`nesting import` whose :s:`CommonPathPrefix` is
+present.
 
 :dp:`fls_UZHHtqJ0ekju`
 An empty :t:`import path prefix` that selects the :t:`entity` of a :t:`simple import` resolves to the current :t:`module`.
--- a/src/entities-and-resolution.rst
+++ b/src/entities-and-resolution.rst
@@ -1166,10 +1166,7 @@ Shadowing
 .. rubric:: Legality Rules
 
 :dp:`fls_ob0riinmitkl`
-:t:`Shadowing` is a property of :t:`[name]s`. A :t:`name` is said to be
-:t:`shadowed` when another :t:`name` with the same characters is introduced
-in the same :t:`scope` within the same :t:`namespace`, effectively hiding it.
-A :t:`name` cannot be referred to by any means once it is :t:`shadowed`.
+:t:`Shadowing` is a property of :t:`[name]s`. A :t:`name` is said to be :t:`shadowed` when another :t:`name` with the same characters is introduced in the same :t:`scope` within the same :t:`namespace`, effectively hiding it. A :t:`name` cannot be referred to once it is :t:`shadowed`, except that a :t:`name` introduced by the :t:`external prelude` may be used as the leftmost :t:`path segment` of a :t:`global path`.
 
 :dp:`fls_fslg89a70e3n`
 No :t:`name` shall be :t:`shadowed` except for
@@ -1611,8 +1608,7 @@ The resolution of the rightmost :t:`path segment` is determined based on the
 :t:`candidate selected entity` is restricted by the :t:`namespace context`.
 
 :dp:`fls_bATFGtxjKq0B`
-It is a static error if the leftmost :t:`path segment` is an :t:`identifier`
-introduced by the :t:`external prelude` that is also :t:`shadowed <shadowing>`.
+It is a static error if the leftmost :t:`path segment` of a :t:`path` that is not a :t:`global path` is an :t:`identifier` introduced by the :t:`external prelude` that is also :t:`shadowed <shadowing>`.
 
 .. _fls_bbso3c45kr9z:
 
--- a/src/entities-and-resolution.rst
+++ b/src/entities-and-resolution.rst
@@ -1005,7 +1005,8 @@ Use Imports
        SimplePath Renaming?
 
    CommonPathPrefix ::=
-       SimplePath? $$::$$
+       SimplePath $$::$$
+     | $$::$$
 
    UseImportContentList ::=
        UseImportContent ($$,$$ UseImportContent)* $$,$$?
@@ -1018,35 +1019,28 @@ A :t:`use import` brings :t:`entities <entity>` :t:`in scope` within the
 :t:`use import` resides.
 
 :dp:`fls_sxo1jb25pl8a`
-A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
-or a :t:`nesting import`.
+A :dt:`common path prefix` is the :t:`simple path` of a :s:`CommonPathPrefix` when one is present, and otherwise consists only of :t:`namespace qualifier` ``::``.
 
 :dp:`fls_WAA4WmohGu6T`
-An :dt:`import path prefix` is the fully constructed :t:`path` prefix of a
-:t:`use import`. An :t:`import path prefix` for a given
-:t:`simple import` or :t:`glob import` is constructed as follows:
+An :dt:`import path prefix` is the fully constructed sequence of :t:`[path segment]s` and :t:`[namespace qualifier]s` that precedes the selected :t:`entity` of a :t:`simple import` or the character asterisk ``*`` of a :t:`glob import`. The :t:`import path prefix` for a given :t:`simple import` or :t:`glob import` is constructed as follows:
 
 #. :dp:`fls_IPYvldMqduf4`
-   Start the :t:`import path prefix` as follows:
+   Make the given :t:`use import` the current :t:`use import`, and start the :t:`import path prefix` as follows:
 
    * :dp:`fls_MOXId37fcNPY`
-     If the :t:`use import` is a :t:`simple import` then start with the
-     :t:`[path segment]s` of the :t:`simple import`'s :t:`simple path`
-     :t:`path prefix`.
+     If the :t:`use import` is a :t:`simple import`, then start with the :t:`simple import`'s :t:`simple path` :t:`path prefix`, retaining any leading :t:`namespace qualifier` of the :t:`simple path`.
 
    * :dp:`fls_2UyFcB6Our1v`
-     If the :t:`use import` is a :t:`glob import` then start with the
-     :t:`[path segment]s` of the :t:`glob import`'s :t:`common path prefix`.
-
-   * :dp:`fls_irdKqoYzBM0M`
-     If the :t:`use import` is a :t:`nesting import` then start with the
-     :t:`[path segment]s` of the :t:`nesting import`'s :t:`common path prefix`.
+     If the :t:`use import` is a :t:`glob import`, then start with the :t:`glob import`'s :t:`common path prefix`, or an empty sequence if the :t:`glob import` lacks a :t:`common path prefix`.
 
 #. :dp:`fls_gAWsqibl4GLq`
-   Then if the current :t:`use import` is the child of a :t:`nesting import`,
-   prepend the :t:`nesting import`'s :t:`common path prefix` to the
-   :t:`import path prefix`. Repeat this step with the :t:`nesting import` as
-   the current :t:`use import`.
+   While the current :t:`use import` is the child of a :t:`nesting import`, repeat the following steps:
+
+   * :dp:`fls_irdKqoYzBM0M`
+     If the :t:`nesting import` has a :t:`common path prefix`, prepend the :t:`common path prefix` to the :t:`import path prefix`. If the :t:`common path prefix` contains a :t:`path segment` and the :t:`import path prefix` was not empty, place a :t:`namespace qualifier` ``::`` between them.
+
+   * :dp:`fls_lfsIPXV0OTI5`
+     Make the :t:`nesting import` the current :t:`use import`.
 
 :dp:`fls_2bkcn83smy2y`
 A :dt:`simple import` is a :t:`use import` that brings into :t:`scope` an :t:`entity` selected by its :t:`simple import path`, or by its :t:`import path prefix` when its :t:`simple path` ends in :t:`keyword` ``self`` and the :t:`simple path` appears in a :t:`nesting import`.
@@ -1055,6 +1049,12 @@ A :dt:`simple import` is a :t:`use import` that brings into :t:`scope` an :t:`en
 A :t:`glob import` is a :t:`use import` that brings all :t:`entities <entity>`
 exported by the :t:`module` or :t:`enum` its :t:`import path prefix` resolves to
 into :t:`scope`.
+
+:dp:`fls_BMtRtjJ7gBKT`
+Each :t:`namespace qualifier` of an :t:`import path prefix` shall either be the first element of the :t:`import path prefix` or appear between two :t:`[path segment]s`.
+
+:dp:`fls_UZHHtqJ0ekju`
+An empty :t:`import path prefix` resolves to the current :t:`module`.
 
 :dp:`fls_JHU0ersYB6eL`
 An :t:`import path prefix` shall resolve to a :t:`module` or :t:`enum`.
@@ -1125,7 +1125,7 @@ A :t:`use import` with a single :t:`path segment` expressed as either :t:`keywor
 When a :t:`path segment` expressed as :t:`keyword` ``super`` is used to import a parent :t:`module`, the imported :t:`entity` shall be subject to a :t:`renaming`.
 
 :dp:`fls_aam34hsRmKU2`
-A :t:`simple import` whose :t:`import path prefix` consists only of :t:`namespace qualifier` ``::`` and whose :t:`simple path` consists of a single :t:`path segment` expressed as :t:`keyword` ``self`` shall not be used.
+An :t:`import path prefix` shall not consist only of :t:`namespace qualifier` ``::`` when it is the :t:`import path prefix` of a :t:`glob import` or selects the :t:`entity` of a :t:`simple import`.
 
 :dp:`fls_LV94x3HlpBWk`
 A :t:`simple import` shall not refer to :t:`[enum variant]s` through a :t:`type alias`.
--- a/src/entities-and-resolution.rst
+++ b/src/entities-and-resolution.rst
@@ -1025,16 +1025,16 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
 or a :t:`nesting import`.
 
 :dp:`fls_WAA4WmohGu6T`
-An :dt:`import path prefix` is a sequence of :t:`[path segment]s` and
-:t:`[namespace qualifier]s` associated with a :t:`simple import` or
-:t:`glob import`. The :t:`import path prefix` is formed by concatenating the
+An :dt:`import path prefix` is a sequence of :t:`[namespace qualifier]s` and
+:t:`[path segment]s` associated with a :t:`glob import` or
+:t:`simple import`. The :t:`import path prefix` is formed by concatenating the
 following sequences in order. When concatenation would place two
 :t:`[path segment]s` next to each other, :t:`namespace qualifier` ``::`` is
 placed between them.
 
 #. :dp:`fls_gAWsqibl4GLq`
-   For each :t:`nesting import` in which the :t:`simple import` or
-   :t:`glob import` is nested, from outermost to innermost:
+   For each :t:`nesting import` in which the :t:`glob import` or
+   :t:`simple import` is nested, from outermost to innermost:
 
    * :dp:`fls_irdKqoYzBM0M`
      The :t:`nesting import`'s :t:`common path prefix` if it has one,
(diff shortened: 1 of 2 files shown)


PLeVasseur:
Me, realizing I did not request @tshepang and @kirtchev-adacore to review 🙃

src/entities-and-resolution.rst
```diff
@@ -1005,7 +1005,8 @@ Use Imports
        SimplePath Renaming?
 
    CommonPathPrefix ::=
```
tshepang, line 1007: looks like this change is not needed since these mean same thing... `SimplePath?` means SimplePath is optional

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/b26f6f6a0b0594afba7fccabc61234ebec212bfb..8e9c110f9933524c092b73c0402d61dcb109955e)*

src/entities-and-resolution.rst
```diff
@@ -1018,35 +1019,28 @@ A :t:`use import` brings :t:`entities <entity>` :t:`in scope` within the
 :t:`use import` resides.
 
 :dp:`fls_sxo1jb25pl8a`
```
tshepang, line 1021: I think we should not consider empty segments as what can be a common path prefix. This makes us adjust text elsewhere without adding clarity. To demo my suggestion, in `::foo::{self, b as foo, c::*}`, `::foo` is the common path prefix. Also, that is equivalent to `::{foo::{self, b as foo, c::*}}` since the outer braces do not add any semantic meaning... they are simply discarded.

Following my suggestion also aligns us more with Reference... its definition of common prefix does not include anything with an empty segment.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/b26f6f6a0b0594afba7fccabc61234ebec212bfb..8e9c110f9933524c092b73c0402d61dcb109955e)*

tshepang (changes requested):

src/entities-and-resolution.rst
kirtchev-adacore: I second that.

src/entities-and-resolution.rst
```diff
@@ -1018,35 +1019,28 @@ A :t:`use import` brings :t:`entities <entity>` :t:`in scope` within the
...
 
 :dp:`fls_WAA4WmohGu6T`
-An :dt:`import path prefix` is the fully constructed :t:`path` prefix of a
-:t:`use import`. An :t:`import path prefix` for a given
-:t:`simple import` or :t:`glob import` is constructed as follows:
+An :dt:`import path prefix` is a sequence of :t:`[path segment]s` and :t:`[namespace qualifier]s` associated with a :t:`simple import` or :t:`glob import`. The :t:`import path prefix` for a given :t:`simple import` or :t:`glob import` is constructed as follows:
 
 #. :dp:`fls_IPYvldMqduf4`
-   Start the :t:`import path prefix` as follows:
+   Make the given :t:`use import` the current :t:`use import`, and start the :t:`import path prefix` as follows:
```
kirtchev-adacore, line 1028: This, coupled with "Make the :t:`nesting import` the current :t:`use import`.", is not correct, and is confusing.

Syntactically, a "use import" is just `use ...;` (see the grammar for `UseImport`). These two rules somewhat imply that the construction of the import path prefix spans across multiple use imports (and I mean multiple **items**), which is just not true.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/b26f6f6a0b0594afba7fccabc61234ebec212bfb..8e9c110f9933524c092b73c0402d61dcb109955e)*

src/entities-and-resolution.rst
```diff
@@ -1018,35 +1019,28 @@ A :t:`use import` brings :t:`entities <entity>` :t:`in scope` within the
...
-     :t:`[path segment]s` of the :t:`glob import`'s :t:`common path prefix`.
+     If the :t:`use import` is a :t:`glob import`, then start with the :t:`glob import`'s :t:`common path prefix`, or an empty sequence if the :t:`glob import` lacks a :t:`common path prefix`.
+
+#. :dp:`fls_gAWsqibl4GLq`
+   While the current :t:`use import` is the child of a :t:`nesting import`, repeat the following steps:
 
    * :dp:`fls_irdKqoYzBM0M`
-     If the :t:`use import` is a :t:`nesting import` then start with the
-     :t:`[path segment]s` of the :t:`nesting import`'s :t:`common path prefix`.
+     If the :t:`nesting import` has a :t:`common path prefix`, prepend the :t:`common path prefix` to the :t:`import path prefix`. If the :t:`common path prefix` contains a :t:`path segment` and the :t:`import path prefix` was not empty, place a :t:`namespace qualifier` ``::`` between them.
```
kirtchev-adacore, line 1040: I see what you are trying to do here, however this is becoming quite confusing. I think we have a higher-level problem with our approach. Perhaps going start-to-end is not the correct approach.

We basically want to identify the prefix that is supposed to resolve to something, and that something depends on the path segments involved, and the use of certain symbols and keywords. For example, in
```rust
use aaa::bbb::{self, ccc::*, ddd::{self, eee}};
```
the prefixes of interest are
- `aaa::bbb::` for the subsequent `self`, `ccc` and `ddd`,
- `aaa::bbb::ccc::` for the subsequent `*`,
- `aaa::bbb::ccc::ddd::` for the subsequent `self` and `eee`.

Perhaps the correct approach is to define "import path prefix" in terms of a end-to-start order, relative to some path segment, `*`, or keyword.

What do you think?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/b26f6f6a0b0594afba7fccabc61234ebec212bfb..8e9c110f9933524c092b73c0402d61dcb109955e)*

src/entities-and-resolution.rst
PLeVasseur: Makes sense, yep.

`SimplePath? ::` already covers the bare `::` form, so the split added no expressiveness.

I restored the original grammar and removed the syntax change from the changelog.

src/entities-and-resolution.rst
PLeVasseur: Got it.

I restored the definition: a common path prefix is the leading simple path. That includes `::foo`, but not bare `::`. The import path prefix construction handles bare `::` separately.

src/entities-and-resolution.rst
PLeVasseur: Agree with you. I removed the mutable "current use import" traversal entirely.

The construction now refers only to the nesting imports that contain the target simple or glob import, followed by that target import's own contribution, so it no longer suggests traversal across `use` items.

src/entities-and-resolution.rst
PLeVasseur: I'm glad you saw the vision, even if it was kind of confusing.

The definition now starts with a particular simple or glob import and identifies its enclosing nesting imports. Outermost-to-innermost is the order of concatenation; there is no mutable traversal.

For your example, the prefixes are `aaa::bbb`, `aaa::bbb::ccc`, and `aaa::bbb::ddd` (`ddd` is a sibling of `ccc`).

PLeVasseur (commented):
Following up! I think this is probably worth another review now.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..60e7f269a66dadc6c9d1daff864837c4a49dc0ba)*

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,22 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
-   Start the :t:`import path prefix` as follows:
+#. :dp:`fls_gAWsqibl4GLq`
+   For each :t:`nesting import` in which the :t:`glob import` or :t:`simple import` is nested, from outermost to innermost:
 
-   * :dp:`fls_MOXId37fcNPY`
-     If the :t:`use import` is a :t:`simple import` then start with the
-     :t:`[path segment]s` of the :t:`simple import`'s :t:`simple path`
-     :t:`path prefix`.
+   * :dp:`fls_irdKqoYzBM0M`
+     The :t:`nesting import`'s :t:`common path prefix` if it has one, :t:`namespace qualifier` ``::`` if it lacks a :t:`common path prefix` and its :s:`CommonPathPrefix` is present, or an empty sequence otherwise.
```
kirtchev-adacore: I could not decipher this paragraphs. The options are
1. The nesting import's common path prefix if it has one
2. :: if it lacks a common path prefix and its CommonPathPrefix is present
3. empty sequence

For 2, what do the two "it"s refer to? If the first "it" refers to the nesting import, then the second "it" cannot be the same nesting import as this would be nonsense.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..60e7f269a66dadc6c9d1daff864837c4a49dc0ba)*

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,22 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
-     If the :t:`use import` is a :t:`glob import` then start with the
-     :t:`[path segment]s` of the :t:`glob import`'s :t:`common path prefix`.
+#. :dp:`fls_IPYvldMqduf4`
+   A sequence determined by the kind of import:
 
-   * :dp:`fls_irdKqoYzBM0M`
-     If the :t:`use import` is a :t:`nesting import` then start with the
-     :t:`[path segment]s` of the :t:`nesting import`'s :t:`common path prefix`.
+   * :dp:`fls_2UyFcB6Our1v`
+     For a :t:`glob import`, the :t:`glob import`'s :t:`common path prefix` if it has one, :t:`namespace qualifier` ``::`` if it lacks a :t:`common path prefix` and its :s:`CommonPathPrefix` is present, or an empty sequence otherwise.
```
kirtchev-adacore: The same issue with the two "it"s is also present here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..60e7f269a66dadc6c9d1daff864837c4a49dc0ba)*

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,22 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
 
-#. :dp:`fls_gAWsqibl4GLq`
-   Then if the current :t:`use import` is the child of a :t:`nesting import`,
-   prepend the :t:`nesting import`'s :t:`common path prefix` to the
-   :t:`import path prefix`. Repeat this step with the :t:`nesting import` as
-   the current :t:`use import`.
+   * :dp:`fls_MOXId37fcNPY`
+     For a :t:`simple import`, the :t:`simple import`'s :t:`simple path` after removing its last :t:`path segment` and, if another :t:`path segment` precedes the last :t:`path segment`, the :t:`namespace qualifier` that separates them.
```
kirtchev-adacore: I think that "and, if another ..." is redundant. If the path is `aaa::bbb`, then after you remove `bbb`, you are left with `aaa::`, so there is no need to mention that the path also retains the namespace qualifier.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..60e7f269a66dadc6c9d1daff864837c4a49dc0ba)*

src/entities-and-resolution.rst
```diff
@@ -1059,8 +1050,14 @@ A :t:`glob import` is a :t:`use import` that brings all :t:`entities <entity>`
 exported by the :t:`module` or :t:`enum` its :t:`import path prefix` resolves to
 into :t:`scope`.
 
+:dp:`fls_BMtRtjJ7gBKT`
+A :t:`glob import`, :t:`nesting import`, or :t:`simple import` expressed with a leading :t:`namespace qualifier` ``::`` shall not be nested, directly or indirectly, within any :t:`nesting import` whose :s:`CommonPathPrefix` is present.
```
kirtchev-adacore: I am not sure what "indirect nesting" is. If you have `aaa::{::bbb}`, then `::bbb` is just "nested". I don't understand the distinction between direct and indirect nesting.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..60e7f269a66dadc6c9d1daff864837c4a49dc0ba)*

src/entities-and-resolution.rst
```diff
@@ -1059,8 +1050,14 @@ A :t:`glob import` is a :t:`use import` that brings all :t:`entities <entity>`
...
+
+:dp:`fls_UZHHtqJ0ekju`
+An empty :t:`import path prefix` that selects the :t:`entity` of a :t:`simple import` resolves to the current :t:`module`.
```
kirtchev-adacore: How about

A simple import with an empty import path prefix resolves to the current module.

This is consistent with how other similar rules are expressed, plus I am not a fan of "selects". Use imports do not "choose", they simply "import" names.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..60e7f269a66dadc6c9d1daff864837c4a49dc0ba)*

src/entities-and-resolution.rst
```diff
@@ -1106,8 +1103,7 @@ A :t:`nesting import` is a :t:`use import` that provides a common
 :t:`common path prefix` for its nested :t:`[use import]s`.
 
 :dp:`fls_iNUBX5fJAI1N`
-A :t:`glob import` outside of a :t:`nesting import` without a :t:`common path
-prefix` is rejected, but may still be consumed by :t:`[macro]s`.
+A :t:`glob import` whose :t:`import path prefix` is empty or consists only of :t:`namespace qualifier` ``::`` is rejected, but may still be consumed by :t:`[macro]s`.
```
kirtchev-adacore: I know you did not write this, but it might be a good opportunity to fix the language here.

The FLS does not use the verb "to reject" because this implies a particular behavior of a conforming tool (ex: The compiler rejects yada yada with an error message). The more neutral way of expressing this scenario is "static error".

So how about

It is a static error if a glob import whose import path prefix is empty or consists only of namespace qualifier ::.

I am not sure whether it is worth mentioning the "consumption by macros" part, as I suspect more illegal constructs can be triaged by macros, but we never mention that interplay.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..60e7f269a66dadc6c9d1daff864837c4a49dc0ba)*

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,22 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
+   * :dp:`fls_2UyFcB6Our1v`
+     For a :t:`glob import`, the :s:`SimplePath` of the :t:`glob import`'s :s:`CommonPathPrefix` if that :s:`SimplePath` is present, :t:`namespace qualifier` ``::`` if the :s:`CommonPathPrefix` is present without a :s:`SimplePath`, or an empty sequence if the :s:`CommonPathPrefix` is absent.
 
-#. :dp:`fls_gAWsqibl4GLq`
-   Then if the current :t:`use import` is the child of a :t:`nesting import`,
-   prepend the :t:`nesting import`'s :t:`common path prefix` to the
-   :t:`import path prefix`. Repeat this step with the :t:`nesting import` as
-   the current :t:`use import`.
+   * :dp:`fls_MOXId37fcNPY`
+     For a :t:`simple import`, the :t:`simple import`'s :t:`simple path` after removing the last :t:`path segment` and any :t:`namespace qualifier` ``::`` that separates the last :t:`path segment` from a preceding :t:`path segment`.
```
kirtchev-adacore, line 1043: I am noticing an inconsistency among the sequence-related rules. Some opt to use syntactic categories (`SimplePath`, `CommonPathPrefix`) while some use terms (`simple path`).

I think using terms instead of syntactic categories would be more appropriate.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/a70c80edeeb38e20dea1040a5773705e3eeeab85..dd52e9df04edd4f3adb927cfe95cb9f601138e82)*

PLeVasseur: Makes sense. The nesting and glob contributions now use the terms throughout, with each case in its own paragraph: the common path prefix if there is one, otherwise a bare leading `::`, otherwise an empty sequence. Same results as before, without `SimplePath` or `CommonPathPrefix` in those rules.

I gave the nesting restriction (`fls_BMtRtjJ7gBKT`) the same treatment too. An enclosing group counts if it has a common path prefix or a leading `::`, so bare-root ancestors are still caught.

PLeVasseur: Makes sense. Have updated to use the terms here and elsewhere.

kirtchev-adacore: The change looks good.

kirtchev-adacore (approved):

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,40 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
-   Start the :t:`import path prefix` as follows:
+#. :dp:`fls_gAWsqibl4GLq`
+   For each :t:`nesting import` in which the :t:`glob import` or :t:`simple import` is nested, from outermost to innermost:
 
-   * :dp:`fls_MOXId37fcNPY`
-     If the :t:`use import` is a :t:`simple import` then start with the
-     :t:`[path segment]s` of the :t:`simple import`'s :t:`simple path`
-     :t:`path prefix`.
+   * :dp:`fls_irdKqoYzBM0M`
+     A sequence determined as follows:
```
tshepang: - looks like the extra paragraph is not needed here
- this also fix grammar
```suggestion
   For each :t:`nesting import` in which the :t:`glob import` or :t:`simple import` is nested, from outermost to innermost, the sequence is determined as follows:
```

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,40 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
+
+     * :dp:`fls_H1hL5ogQjruY`
+       If the :t:`nesting import` has a :t:`common path prefix`, then the sequence is that :t:`common path prefix`.
+
+     * :dp:`fls_ZTDNlERRToGG`
+       Otherwise, if the :t:`nesting import` is expressed with a leading :t:`namespace qualifier` ``::``, then the sequence consists only of that :t:`namespace qualifier`.
+
+     * :dp:`fls_9dRQ4YGKstSN`
+       Otherwise the sequence is empty.
+
```
tshepang: this seems invalid, since a nesting import has a common path prefix or not
but maybe am failing to follow reasoning
```suggestion
```

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,40 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
-   Then if the current :t:`use import` is the child of a :t:`nesting import`,
-   prepend the :t:`nesting import`'s :t:`common path prefix` to the
-   :t:`import path prefix`. Repeat this step with the :t:`nesting import` as
-   the current :t:`use import`.
+     * :dp:`fls_zth1OR1dU4q6`
+       Otherwise, if the :t:`glob import` is expressed with a leading :t:`namespace qualifier` ``::``, then the sequence consists only of that :t:`namespace qualifier`.
+
+     * :dp:`fls_GO9rlHrbXAZn`
+       Otherwise the sequence is empty.
+
```
tshepang: similar to https://github.com/rust-lang/fls/pull/717/changes#r4092739338
```suggestion
```

src/entities-and-resolution.rst
```diff
@@ -1025,31 +1025,40 @@ A :dt:`common path prefix` is the leading :t:`simple path` of a :t:`glob import`
...
+   * :dp:`fls_MOXId37fcNPY`
+     For a :t:`simple import`, the :t:`simple import`'s :t:`simple path` after removing the last :t:`path segment` and any :t:`namespace qualifier` ``::`` that separates the last :t:`path segment` from a preceding :t:`path segment`.
```
tshepang: avoids making it seem like the namespace qualifier is optional here
```suggestion
     For a :t:`simple import`, the :t:`simple import`'s :t:`simple path` after removing the last :t:`path segment` and the :t:`namespace qualifier` ``::`` preceding the :t:`path segment`.
```

src/entities-and-resolution.rst
```diff
@@ -1059,8 +1068,14 @@ A :t:`glob import` is a :t:`use import` that brings all :t:`entities <entity>`
 exported by the :t:`module` or :t:`enum` its :t:`import path prefix` resolves to
 into :t:`scope`.
 
+:dp:`fls_BMtRtjJ7gBKT`
+A :t:`glob import`, :t:`nesting import`, or :t:`simple import` expressed with a leading :t:`namespace qualifier` ``::`` shall not be nested within any :t:`nesting import` that has a :t:`common path prefix` or is expressed with a leading :t:`namespace qualifier` ``::``.
```
tshepang: using a shortcut since this is all of them
```suggestion
A :t:`use import` expressed with a leading :t:`namespace qualifier` ``::`` shall not be nested within any :t:`nesting import` that has a :t:`common path prefix` or is expressed with a leading :t:`namespace qualifier` ``::``.
```

tshepang: I understand the part "A use import expressed with a leading namespace qualifier shall not be nested within any nesting import" to mean that `use a::{::b, c};` is not allowed, but I don't understand the rest of the sentence.

src/entities-and-resolution.rst
```diff
@@ -1059,8 +1068,14 @@ A :t:`glob import` is a :t:`use import` that brings all :t:`entities <entity>`
...
+
+:dp:`fls_UZHHtqJ0ekju`
+An empty :t:`import path prefix` of a :t:`simple import` whose :t:`simple path` is expressed as :t:`keyword` ``self`` resolves to the current :t:`module`.
+
```
tshepang: I don't we should have such a thing as empty path prefix 
```suggestion
```

src/entities-and-resolution.rst
```diff
@@ -1059,8 +1068,14 @@ A :t:`glob import` is a :t:`use import` that brings all :t:`entities <entity>`
...
 :dp:`fls_JHU0ersYB6eL`
-An :t:`import path prefix` shall resolve to a :t:`module` or :t:`enum`.
+An :t:`import path prefix` that contains a :t:`path segment` shall resolve to an :t:`enum` or :t:`module`.
```
tshepang: related to https://github.com/rust-lang/fls/pull/717/changes#r4092947510

tshepang (changes requested):
Have left some comments, but maybe before handling those, I do wonder why we need the concept of an "import path prefix" when we have "common path prefix". To support that, I don't think we should treat "global paths" (paths starting with `::`) in such a special way... they should only add additional tooling-related behavior, and the distinction defined in just one place.

With examples, I think in...
- `use a::{b, c};`, `a` should be a "common path prefix"
- `use ::a::{b, c};`, `a` (or `::a`, not sure) should similarly be considered a "common path prefix"
  - additional semantics of this being a "global path" can be explained elsewhere

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/717/0626fa45a0d296e59e62b183465bc7f114d1031e..badc40451e9a3a96ae5dd055e2f0d33b8181f5e4)*

PR #709, opened 2026-07-06
Document if let guards

PLeVasseur:
This is my attempt to carry forward rust-lang/fls#693 and incorporate the suggestions and feedback received there.

It documents `if let` guards on match arms as stabilized in Rust 1.95.0, covering the affected syntax, typing, scoping, evaluation, exhaustiveness, and destructor rules.

PR #718, opened 2026-08-17
some cleaning

tshepang:
found these easy ones from rust-lang/fls#709

### DIFF
--- a/src/changelog.rst
+++ b/src/changelog.rst
@@ -126,7 +126,6 @@ Language changes in Rust 1.95.0
   - :p:`fls_Gc5RpT8nYvQm`
   - :p:`fls_Nw2KxL7qVmRs`
   - :p:`fls_Rh9TsD4vQpLm`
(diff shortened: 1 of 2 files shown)


src/types-and-traits.rst
```diff
@@ -939,8 +939,6 @@ A :t:`function item type` implements the :std:`core::ops::Fn` :t:`trait`, the :s
 Indirection Types
 -----------------
 
-.. rubric:: Legality Rules
```
kirtchev-adacore, line 942: I think this is needed. Definitions are still considered legality rules.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/fls/718/084dc919b12d90379df78e19d5a831bd879a4360..c532b5e19dd63530da7ec8b88f6a4bfd027137d3)*

tshepang: oh, this is not intended

kirtchev-adacore (approved):

